PR #54, opened 2021-03-24
Fix read logic issue

alexr-bq:
### Summary

Fixed read logic to properly read multiple file ranges in a partition.

### Description

There was a logical issue in the read pipe, where it expected a "Done Reading" Error type when a file range was done reading.  It would instead simply receive an empty data block. The logic for setting a flag and then returning a "done reading" error is now eliminated.

Also made a change so that if cleanup files get left over due to some problem, they will not cause an issue for the next read attempt.

### Related Issue

VER-76600

### Additional Reviewers

@jonathanl-bq 
@NerdLogic

### DIFF
--- a/connector/src/test/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemReadPipeTests.scala
+++ b/connector/src/test/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemReadPipeTests.scala
@@ -554,7 +554,7 @@ class VerticaDistributedFilesystemReadPipeTests extends AnyFlatSpec with BeforeA
     }
 
     pipe.readData match {
-      case Left(_) => fail
+      case Left(err) => fail(err.getFullContext)
       case Right(data) =>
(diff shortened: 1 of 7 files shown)


connector/src/test/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemReadPipeTests.scala
```diff
@@ -554,8 +554,9 @@ class VerticaDistributedFilesystemReadPipeTests extends AnyFlatSpec with BeforeA
     }
 
     pipe.readData match {
-      case Left(err) => assert(err.getError == DoneReading())
-      case Right(_) => fail
+      case Left(_) => fail
```
jonathanl-bq, line 557: fail(err.getFullContext)

jonathanl-bq (approved):
One small comment, but looks fine

NerdLogic:
Alex, please click that CLA Assistant button in the comments.

PR #64, opened 2021-04-13
VER-76705: Remove log level option from the connector

jonathanl-bq:
### Summary

This change removes the log level user option from the connector, since it's not needed.

### Description

The connector uses Spark's log4j.properties file to configure logging, so this option wasn't needed. This change removes that option and updates our documentation to remove the option and add a section on configuring logging.

### Related Issue

http://jira.verticacorp.com:8080/jira/browse/VER-76705

### Additional Reviewers

@NerdLogic 
@alexr-bq


connector/src/main/scala/com/vertica/spark/config/LogProvider.scala
```diff
@@ -0,0 +1,28 @@
...
+import com.typesafe.scalalogging.Logger
+
+/**
+ * Used to provide a logger for a given class, configured with a given log level.
+ */
+case object LogProvider {
+  def getLogger(c: Class[_]): Logger = {
+    val logger = Logger(c)
+
+    logger
```
alexey-temnikov, line 25: nit?: Would it be better to shorten these 3 lines just to `Logger(c)` ?

alexr-bq (approved):

PR #71, opened 2021-04-30
Support kerberized hdfs without vertica

alexr-bq:
### Summary

Support kerberized hdfs without vertica.

### Description

Attempt to set kerberos config if hdfs configuration is set to work with kerberos. 

### Related Issue

VER-76995

### Additional Reviewers

@jonathanl-bq


connector/src/main/scala/com/vertica/spark/datasource/jdbc/VerticaJdbcLayer.scala
```diff
@@ -288,7 +288,7 @@ class VerticaJdbcLayer(cfg: JDBCConfig) extends JdbcLayerInterface {
         val hadoopConf = session.sparkContext.hadoopConfiguration
         val isKerberosEnabled = hadoopConf.get("hadoop.security.authentication")
         logger.debug("Hadoop impersonation: kerb authentication: " + isKerberosEnabled)
-        if (isKerberosEnabled == "kerberos") {
+        if (isKerberosEnabled != null && isKerberosEnabled == "kerberos") {
```
jonathanl-bq, line 291: This variable name makes no sense. It's not a boolean.

jonathanl-bq (approved):
Rename the variable I commented on and it should be fine. Maybe name it something like "authenticationMethod".

PR #76, opened 2021-05-07
Ver 77161: Warn on discarded non-Unit values

jonathanl-bq:
### Summary

This PR makes it so that the compiler produces warnings when non-Unit values are discarded. This should make error handling more robust as unhandled errors cannot be silently ignored.

### Description

The build.sbt file for the main connector source code is updated to enable warning on discarded non-Unit values. This change also adds an explicit ignore function under general Utils to allow us to explicitly ignore returned values when necessary. In addition, this change fixes several warnings that showed up as a result of enabling this option.

### Related Issue

http://jira.verticacorp.com:8080/jira/browse/VER-77161

### Additional Reviewers

@alexr-bq 
@NerdLogic


connector/src/main/scala/com/vertica/spark/datasource/v2/VerticaDatasourceV2Write.scala
```diff
@@ -168,7 +169,7 @@ class VerticaBatchWriter(config: WriteConfig, writer: DSWriterInterface) extends
   * Initiates final stages of writing for a failed write of this partition.
   */
   override def abort(): Unit = {
-    writer.closeWrite()
+    Utils.ignore(writer.closeWrite())
```
alexr-bq, line 172: Maybe add a comment saying why this is ignored.

alexr-bq (approved):

PR #80, opened 2021-05-14
Change examples to use jar from maven central and test

ravjotbrar:
### Summary

Added library dependency for assembly JAR for connector from maven central.

### Description

When users were running the examples, they had to manually add a lib folder along with the JAR file. This step is now removed by adding the library dependency from maven central in the build.sbt file for each example. This change is also reflected in the READ.md file in the examples folder. 

### Related Issue

http://jira.verticacorp.com:8080/jira/browse/VER-77164

### Additional Reviewers

@alexr-bq 
@jonathanl-bq 
@alexey-temnikov

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -7,23 +7,12 @@ Clone this repository:
 ```
 git clone https://github.com/vertica/spark-connector.git
 ```
-
 From the project's root directory:
 ```
 cd docker
 docker compose up -d
 ```
 This will create a docker image for a client container and docker containers for a sandbox client environment and single-node clusters for both Vertica and HDFS.
-
-Next, change directory to the example you want to run. For example, if you want to run the "demo" example:
-```
-cd examples/demo
-```
-Now create a lib directory:
-```
-mkdir lib
-```
-Put the assembly jar for the connector in the lib folder you just created. If you don't have the jar file, you can download it here: https://github.com/vertica/spark-connector/releases
 
 Run the following commands to update the HDFS configuration files and restart HDFS:
 ```
@@ -35,4 +24,4 @@ docker exec docker_hdfs_1 /opt/hadoop/sbin/start-dfs.sh
 
 Run `docker exec -it docker_client_1 /bin/bash` to enter the sandbox client environment.
 
-Now just run `sbt run` from the `/spark-connector/examples/demo` directory.
+Now just run `sbt "run [CASE]"` from the `/spark-connector/examples/demo` directory.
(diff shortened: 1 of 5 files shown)


examples/demo/build.sbt
```diff
@@ -20,3 +20,4 @@ resolvers += "Artima Maven Repository" at "https://repo.artima.com/releases"
 resolvers += "jitpack" at "https://jitpack.io"
 
 libraryDependencies += "com.typesafe" % "config" % "1.4.1"
+libraryDependencies += "com.vertica.spark" % "vertica-spark" % "2.0.0-rc0"
```
alexr-bq, line 23: May as well make this consistent with the others and use the Seq

alexr-bq (approved):
Looks good! Sign the CLA

alexr-bq:
Just to confirm: you tested this with nothing in lib/ dir?

examples/demo/build.sbt
ravjotbrar: Done

ravjotbrar:
> Just to confirm: you tested this with nothing in lib/ dir?

That's right. No lib directory in any of the examples.

examples/README.md
```diff
@@ -35,4 +24,4 @@ docker exec docker_hdfs_1 /opt/hadoop/sbin/start-dfs.sh
 
 Run `docker exec -it docker_client_1 /bin/bash` to enter the sandbox client environment.
 
-Now just run `sbt run` from the `/spark-connector/examples/demo` directory.
+Now just run `sbt "run [CASE]"` from the `/spark-connector/examples/demo` directory.
```
jonathanl-bq, line 27: We should be more general and state how to run any of the examples, not just the cases from the demo example.

jonathanl-bq (changes requested):

examples/README.md
ravjotbrar: What I'm thinking is to use the basic-write example as a case in the demo example. We already have a basicRead case, so the basic-read example seems redundant. This way, all specific examples/cases are organized in one directory. What do you think?

jonathanl-bq: I'm okay with combining some of the other examples into the demo project, though I'm not sure what @alexr-bq thinks. I'm more concerned that there's now only instructions on how to run examples from the demo project. The new README doesn't explain how to run the Kerberos example, or any other future examples we might add (e.g. S3, TLS, etc.).

ravjotbrar: I'll change that in the ticket that involves updating the documentation of the README.

jonathanl-bq: Okay, fair enough.

jonathanl-bq (approved):

PR #83, opened 2021-05-18
Ver 76342: S3: Add options to connector for access key

jonathanl-bq:
### Summary

This change adds support for connecting to S3. It also exposes a few new AWS connector options.

### Description

Environment variables are read to set the access key ID, secret access key, and region for AWS. Config and error tests are added. I manually tested that all integration tests pass with S3 being used instead of HDFS. Documentation was also updated with instructions on how to use the connector with S3 + new connector options.

### Related Issue

http://jira.verticacorp.com:8080/jira/browse/VER-76342

### Additional Reviewers

@alexr-bq 
@NerdLogic 
@ravjotbrar


connector/src/main/scala/com/vertica/spark/datasource/core/DSConfigSetup.scala
```diff
@@ -154,6 +154,28 @@ object DSConfigSetupUtils {
     }
   }
 
+  def getAWSAuth(config: Map[String, String]): ValidationResult[Option[AWSAuth]] = {
+    (sys.env.get("AWS_ACCESS_KEY_ID"), sys.env.get("AWS_SECRET_ACCESS_KEY")) match {
```
alexr-bq, line 158: I think our config should have priority over env vars

connector/src/main/scala/com/vertica/spark/datasource/core/DSConfigSetup.scala
```diff
@@ -154,6 +154,28 @@ object DSConfigSetupUtils {
...
+      case (Some(accessKeyId), Some(secretAccessKey)) => Some(AWSAuth(accessKeyId, secretAccessKey)).validNec
```
alexr-bq, line 159: Can we also check the spark context for these settings being set that way. Want to cover the bases of the way users might already be setting these for spark.

alexr-bq (changes requested):

README.md
```diff
@@ -110,6 +122,11 @@ Below is a detailed list of connector options:
...
+| aws_access_key_id                              | String                                             | The access key ID for S3. Specifying this option sets the access key ID at the session level. Alternatively, you can set the environment variable AWS_ACCESS_KEY_ID.                                                                                                                                                                                                                                                                                                                                                                                           | No (Yes if aws_secret_access_key is specified)                     |                             |
```
alexr-bq, line 125: Add note to staging_fs_url to use s3a://

alexr-bq (commented):
Add note

connector/src/main/scala/com/vertica/spark/datasource/core/DSConfigSetup.scala
jonathanl-bq: Done, but it did mean bringing SparkSession into our config unit tests.

alexr-bq (approved):

PR #99, opened 2021-06-01
Add fix for issues with pushdown and repeated calls to create scans

alexr-bq:
### Summary

Add fix for issues with pushdown and repeated calls to create scans

### Description

Two issues:
- We had a bug where if two different filters were called and a scan was created for them rapidly enough, it would use the same filter twice
- Now if we detect to be using a new filter we consider it a new operation

### Additional Reviewers

@jonathanl-bq 
@ravjotbrar


functional-tests/build.sbt
```diff
@@ -16,23 +16,25 @@ name := "spark-vertica-connector-functional-tests"
...
 libraryDependencies += "com.typesafe" % "config" % "1.4.1"
 
 libraryDependencies += "org.scala-lang.modules" %% "scala-parser-combinators" % "1.1.2"
-libraryDependencies += "com.vertica.jdbc" % "vertica-jdbc" % "10.0.1-0"
-libraryDependencies += "org.apache.spark" %% "spark-core" % "3.0.0"
-libraryDependencies += "org.apache.spark" %% "spark-sql" % "3.0.0"
-libraryDependencies += "org.apache.hadoop" % "hadoop-hdfs" % "2.4.0"
+libraryDependencies += "org.apache.spark" %% "spark-core" % sparkVersion
+libraryDependencies += "org.apache.spark" %% "spark-sql" % sparkVersion
+libraryDependencies += "org.apache.hadoop" % "hadoop-hdfs" % "3.3.0"
```
alexey-temnikov, line 30: Will it break test execution locally when we are using sbt run?

alexr-bq: Apparently so.

jonathanl-bq (approved):

PR #100, opened 2021-06-01
Ver 77448: Spark timestamp issue returned (on Spark 3.1 only?)

jonathanl-bq:
### Summary

This change adds support for Spark 3.1, by fixing an issue with legacy timestamps.

### Description

Included in this change are new copies of the Spark ReadSupport related files for Spark 3.1. Spark 3.1.0 adds the LEGACY_PARQUET_INT96_REBASE_MODE_IN_WRITE property, so this change sets that property in the FS Write Pipe and FS layer. This change also upgrades our connector and functional test dependencies to use spark 3.1.1 and hadoop-hdfs 3.3.0. The README is also updated to reflect the new dependency on Hadoop 3.3.0 required when using Spark 3.1.1.

### Related Issue

http://jira.verticacorp.com:8080/jira/browse/VER-77448

### Additional Reviewers

@NerdLogic 
@alexr-bq 
@ravjotbrar


functional-tests/src/main/scala/Main.scala
```diff
@@ -164,6 +164,7 @@ object Main extends App {
                                                              awsRegion,
                                                              awsSessionToken,
                                                              awsCredentialsProvider))
+  /*
```
alexr-bq, line 167: Remove commented?

alexr-bq (changes requested):

alexr-bq (approved):

functional-tests/src/main/scala/Main.scala
jonathanl-bq: Done.

PR #102, opened 2021-06-02
Add endpoint parameter to connector

alexr-bq:
### Summary

Adds endpoint and enable ssl parameters to s3 support in the connector.

### Description

The connector may be used with other data sources that implement the S3 API. The endpoint param is used to specify which address, and the enable ssl param is used if the endpoint in question is not https.

### Related Issue

VER-77415

### Additional Reviewers

@jonathanl-bq 
@ravjotbrar


alexey-temnikov (approved):

connector/src/main/scala/com/vertica/spark/datasource/fs/FileStoreLayerInterface.scala
```diff
@@ -185,6 +187,19 @@ class HadoopFileStoreLayer(fileStoreConfig : FileStoreConfig, schema: Option[Str
...
+    case Some(endpoint) =>
+      hdfsConfig.set(S3_ENDPOINT, endpoint.arg)
+      logger.info(s"Loaded $S3_ENDPOINT from ${endpoint.origin}")
+    case None => logger.debug("Did not set AWS endpoint, using default.")
+  }
+  awsOptions.enableSSL match {
+    case Some(enable) =>
+      hdfsConfig.set(S3_ENABLE_SSL, enable.arg)
+      logger.info(s"Loaded $S3_ENABLE_SSL from ${enable.origin}")
+    case None => logger.debug("Did not set AWS endpoint SSL enabling, using default of true.")
```
jonathanl-bq, line 200: Is this string correct? It's a little confusing to read.

jonathanl-bq (approved):
One comment, otherwise this looks fine.

PR #105, opened 2021-06-02
Update documentation and create script for s3-example

ravjotbrar:
### Summary

Updated documentation for s3-example so users do not try and compile inside docker container. Also, created a script for S3 setup.

### Description

Removed documentation that asked user to create a lib directory for the JAR inside s3-example. Also, asked users to run 'sbt assembly' before entering the sandbox environment in docker. This speeds up the compile time and avoids memory issues. In addition, created a script which involves all of the setup required for s3. This simplifies running the s3 example. 

### Related Issue

http://jira.verticacorp.com:8080/jira/browse/VER-77204

### Additional Reviewers

@jonathanl-bq 
@alexr-bq

### DIFF
--- a/examples/s3-example/build.sbt
+++ b/examples/s3-example/build.sbt
@@ -24,7 +24,7 @@ libraryDependencies ++= Seq(
   "org.apache.spark" %% "spark-core" % "3.0.0",
   "org.apache.spark" %% "spark-sql" % "3.0.0",
   "org.apache.hadoop" % "hadoop-aws" % "3.3.0",
-  "com.vertica.spark" % "vertica-spark" % "2.0.0-rc0"
+  "com.vertica.spark" % "vertica-spark" % "2.0.0-0"
 )
 
 assemblyMergeStrategy in assembly := {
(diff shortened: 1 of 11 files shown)


examples/s3-example/build.sbt
```diff
@@ -12,7 +12,7 @@
 // limitations under the License.
 
 scalaVersion := "2.12.12"
-name := "spark-vertica-connector-functional-tests"
+name := "spark-vertica-connector-s3-example"
```
jonathanl-bq, line 15: Shouldn't we be adding the jar as a library dependency here? We're removing the lib folder instructions, so I don't know where this example would get its jar from.

jonathanl-bq (changes requested):

examples/s3-example/build.sbt
ravjotbrar: Added

alexr-bq (approved):

jonathanl-bq (approved):

PR #108, opened 2021-06-04
Run different spark version integration tests

alexr-bq:
### Summary

Run different spark version integration tests

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -59,7 +59,9 @@ jobs:
           path: ./functional-tests/lib/
       - name: Copy functional tests to home directory of client container
         run: docker exec docker_client_1 cp -r /spark-connector/functional-tests /home
-      - name: Run the integration tests
-        run: docker exec -w /home/functional-tests docker_client_1 sbt run
+      - name: Run the integration tests on Spark 3.0
+        run: docker exec -w /home/functional-tests docker_client_1 sbt run -DsparkVersion="3.0.2"
+      - name: Run the integration tests on Spark 3.1
+        run: docker exec -w /home/functional-tests docker_client_1 sbt run -DsparkVersion="3.1.1"
       - name: Remove docker containers
         run: cd docker && docker-compose down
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -59,7 +59,9 @@ jobs:
           path: ./functional-tests/lib/
       - name: Copy functional tests to home directory of client container
         run: docker exec docker_client_1 cp -r /spark-connector/functional-tests /home
-      - name: Run the integration tests
-        run: docker exec -w /home/functional-tests docker_client_1 sbt run
+      - name: Run the integration tests on Spark 3.0
```
jonathanl-bq, line 62: We're not going to run them in parallel?

alexr-bq: This was just to test initially. I could set them to run in parallel, my only concern is could we run into any resource limits?

jonathanl-bq: I don't see why we would. Wouldn't running in sequence take longer and be more costly, if anything?

alexr-bq: Maybe, I don't know enough about it, my logic was assuming each independent run uses a VM or something, there might be a limit on those

.github/workflows/main.yml
```diff
@@ -60,8 +60,8 @@ jobs:
       - name: Copy functional tests to home directory of client container
         run: docker exec docker_client_1 cp -r /spark-connector/functional-tests /home
       - name: Run the integration tests on Spark 3.0
-        run: docker exec -w /home/functional-tests docker_client_1 sbt run -DsparkVersion="3.0.2"
+        run: docker exec -w /home/functional-tests docker_client_1 sbt run -DsparkVersion="3.0.2" -DhadoopVersion="2.4.0" 
       - name: Run the integration tests on Spark 3.1
-        run: docker exec -w /home/functional-tests docker_client_1 sbt run -DsparkVersion="3.1.1"
+        run: docker exec -w /home/functional-tests docker_client_1 sbt run -DsparkVersion="3.1.1" -DhadoopVersion="3.3.0
```
ravjotbrar: Is this the right way to pass env variables to a container?

ravjotbrar: Also missing closing double quotes at the end of line 65

alexr-bq: it's passing it to the sbt run

jonathanl-bq (approved):

.github/workflows/main.yml
alexr-bq: Ok so I think this one parallelization is ok but we shouldn't push it much further if there are more configurations that we need to test.

PR #111, opened 2021-06-11
Ensure clean-up is done in functional tests

ravjotbrar:
### Summary

This update ensures cleanup is done in all functional tests, including after calls that only invoked the initial setup of a read. 

### Description

In the HDFS tests, we now clean up the root directory of the bucket, instead of only the dirtest directory by calling removeDir on the filestore config address. Similarly, in the cleanuputils tests, we clean up the root directory instead of only /CleanupTest. Finally, in the EndtoEnd tests, we assert that all files are cleaned up after each test. We manually clean up files in cases that call our initial read setup, but don't actually read the data. In addition, cleanup is now triggered on reads involving empty tables, and the path that is appended to the test bucket in write tests is removed. 

### Related Issue

http://jira.verticacorp.com:8080/jira/browse/VER-77503

### Additional Reviewers

@jonathanl-bq 
@alexr-bq

### DIFF
--- a/connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemReadPipe.scala
+++ b/connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemReadPipe.scala
@@ -284,6 +284,11 @@ class VerticaDistributedFilesystemReadPipe(
       totalRowGroups = fileMetadata.map(_.rowGroupCount).sum
 
       _ = logger.info("Total row groups: " + totalRowGroups)
+      // If table is empty, cleanup
+      _ =  if(totalRowGroups == 0) {
+        logger.info("Cleaning up all files in path: " + hdfsPath)
+        cleanupUtils.cleanupAll(fileStoreLayer, hdfsPath)
+      }
 
       partitionCount = if (totalRowGroups < requestedPartitionCount) {
         logger.info("Less than " + requestedPartitionCount + " partitions required, only using " + totalRowGroups)
--- a/connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemReadPipe.scala
+++ b/connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemReadPipe.scala
@@ -286,7 +286,7 @@ class VerticaDistributedFilesystemReadPipe(
       _ = logger.info("Total row groups: " + totalRowGroups)
       // If table is empty, cleanup
       _ =  if(totalRowGroups == 0) {
-        logger.info("Cleaning up all files in path: " + hdfsPath)
+        logger.debug("Cleaning up empty directory in path: " + hdfsPath)
         cleanupUtils.cleanupAll(fileStoreLayer, hdfsPath)
       }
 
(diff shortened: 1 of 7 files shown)


connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemReadPipe.scala
```diff
@@ -284,6 +284,11 @@ class VerticaDistributedFilesystemReadPipe(
       totalRowGroups = fileMetadata.map(_.rowGroupCount).sum
 
       _ = logger.info("Total row groups: " + totalRowGroups)
+      // If table is empty, cleanup
+      _ =  if(totalRowGroups == 0) {
+        logger.info("Cleaning up all files in path: " + hdfsPath)
```
alexey-temnikov, line 289: Do we need info log? Can we downgrade it to Debug?

functional-tests/src/main/scala/com/vertica/spark/functests/EndToEndTests.scala
```diff
@@ -638,6 +647,9 @@ class EndToEndTests(readOpts: Map[String, String], writeOpts: Map[String, String
     assert(r2 == (n + 1))
 
     stmt.execute("drop table " + tableName1)
+    fsLayer.removeDir(fsConfig.address)
+    // Need to recreate the root directory for the afterEach assertion check
+    fsLayer.createDir(fsConfig.address, "777")
```
alexey-temnikov, line 652: Why directory should be recreated? If it is absolutely required suggesting to extract it in a separate function to avoid copy pasting  in all tests.

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemReadPipe.scala
ravjotbrar: We used log.info when we're cleaning up in other parts of our code, so I was following that. We can downgrade it if others don't mind.

functional-tests/src/main/scala/com/vertica/spark/functests/EndToEndTests.scala
ravjotbrar: The directory needs to be recreated because that command removes the root directory. So when our afterEach logic checks if our directory has files in it, it will fail on path does not exist instead of checking if it's empty. I believe that is only used in 7 test cases, but I can still extract it in a function if that's preferred.

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemReadPipe.scala
alexr-bq: The message isn't really accurate anyway, as all we're cleaning up is the empty directories.

alexr-bq (approved):

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemReadPipe.scala
ravjotbrar: Changed to "Cleaning up empty directory" and set logger type to debug.

PR #116, opened 2021-06-19
Fix scripts for Pyspark and s3 examples

jonathanl-bq:
### Summary

This change updates documentation and scripts for the PySpark example.

### Description

This change adds automation for the PySpark example to make it easier to run. Previously, we didn't install Python3 or setup the Spark cluster, so this makes it easier to get started. It also adds comments to the Python script to make it easier for people unfamiliar with PySpark to understand what the script is doing. In addition, the READMEs for PySpark and for the examples in general have both been updated.

No changes were required for S3, aside from the README. Apparently, the "bad array subscript" issue I get is only happening for me, and only sometimes. I'll likely need to dig into it a bit more, but for now, this should work fine.

### Related Issue

VER-77686

### Additional Reviewers

@alexr-bq 
@ravjotbrar

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -52,7 +52,7 @@ Once you're finished running the examples exit out of the interactive terminal b
 
 ## Running Example Application in Spark Cluster
 
-Note: this method is required for the S3-example. 
+Note: this method is required for the S3-example and PySpark example. 
 
 In the example's root directory (spark-connector/examples/[EXAMPLE]) , run `sbt assembly`:
 ```
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,6 +1,6 @@
 # How to run the examples
 
-**Note**: The instructions for running the S3, Kerberos, and PySpark example are different. If you would like to run that example, please see their respective READMEs. 
+**Note**: The instructions for running the S3, Kerberos, and PySpark example are different. If you would like to run those examples, please see their respective READMEs. 
 
 Make sure you have docker and sbt installed.
 Tested using docker 20.10.0, sbt 1.4.1
@@ -52,7 +52,7 @@ Once you're finished running the examples exit out of the interactive terminal b
 
 ## Running Example Application in Spark Cluster
 
-Note: this method is required for the S3-example and PySpark example. 
+Note: this method is required for the S3-example. 
 
 In the example's root directory (spark-connector/examples/[EXAMPLE]) , run `sbt assembly`:
 ```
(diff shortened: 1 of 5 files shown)


examples/README.md
```diff
@@ -52,7 +52,7 @@ Once you're finished running the examples exit out of the interactive terminal b
 
 ## Running Example Application in Spark Cluster
 
-Note: this method is required for the S3-example. 
+Note: this method is required for the S3-example and PySpark example.
```
ravjotbrar, line 55: Although running the PySpark example is ran in the spark cluster, the first step listed in this method is an `sbt assembly`, which is not applicable. We might want to clarify that in the README.

ravjotbrar: We also use an entirely different script for this example, right?

examples/pyspark/README.md
```diff
@@ -2,4 +2,4 @@
 
 The connector can be used with pyspark, it must simply be sourced as a JAR file. 
 
-An example application is provided alongside a .sh file which shows how to run such an application with the connector.
+An example application is provided alongside a .sh file which shows how to run such an application with the connector. It can be run using "./run-example.sh". This script will download Python3, Spark, and Hadoop and configure them before running the example.
```
ravjotbrar, line 5: ./run-example.sh should be changed to ./run-python-example.sh, unless we want to supply sparkapp.py as an argument and install python3 in our run-example script.

ravjotbrar (changes requested):

examples/s3-example/README.md
```diff
@@ -52,6 +52,6 @@ docker compose up -d
 ```
 Run `docker exec -it docker_client_1 /bin/bash` to enter the sandbox client environment. The following steps assume you are in this container.
```
ravjotbrar, line 54: We should probably add a step here to cd back into the examples directory.

ravjotbrar: Also should we be using sandbox-clientenv.sh instead of manually inputting the docker commands?

examples/README.md
jonathanl-bq: Somehow, I thought this used SBT, but yes, this shouldn't be here.

examples/pyspark/README.md
jonathanl-bq: Good catch. Fixed.

examples/s3-example/README.md
jonathanl-bq: Yes, this was written before the sandbox-clientenv.sh script was merged. I've updated this now.

ravjotbrar (approved):

Issue #118, opened 2021-06-22
Prototype for Merge statements


alexey-temnikov:
![updatedMerge](https://user-images.githubusercontent.com/83892020/123134871-1577d800-d406-11eb-9e57-d230c350b8ed.png)

With the implementation of merge statements in our connector, users can process raw data in Spark and merge that data with an existing table in Vertica. The processed data will be written to a temporary table in Vertica before being merged. In the existing table, matched records will be updated and new records will be inserted, effectively constituting an “Upsert”. In order to execute a merge statement in our Spark Connector, the user needs to pass in a mergeKey, which will likely be an array of column attributes to join the existing table and temporary table on. If this option is not populated, a write will be performed as usual without the use of a temporary table. 

Example use case: https://www.vertica.com/docs/9.2.x/HTML/Content/Authoring/AdministratorsGuide/Tables/MergeTables/MergeExample.htm

PR #123, opened 2021-06-28
Merge statement prototype

ravjotbrar:
### Summary

With the implementation of merge statements in our connector, users can process raw data in Spark and merge that data with an existing table in Vertica. The processed data will be written to a temporary table in Vertica before being merged. In the existing table, matched records will be updated and new records will be inserted, effectively constituting an “Upsert”. In order to execute a merge statement in our Spark Connector, the user needs to pass in a mergeKey, which will likely be an array of column attributes to join the existing table and temporary table on. If this option is not populated, a write will be performed as usual without the use of a temporary table.

### Description

For this prototype, changes were made to add a new connector option, which is used to pass in the merge key. When in the write pipe, the connector checks for the existence of this merge key, and if it exists, creates a temporary table which will be used for the intermediary copy. The write pipe then uses the JDBC layer to perform the merge from the temporary table into the target table. Two new supporting functions were implemented in SchemaTools to create the merge statement.

### Related Issue

https://github.com/vertica/spark-connector/issues/118

### Additional Reviewers

@alexr-bq 
@jonathanl-bq 
@alexey-temnikov

### DIFF
--- a/connector/src/main/scala/com/vertica/spark/config/WriteConfig.scala
+++ b/connector/src/main/scala/com/vertica/spark/config/WriteConfig.scala
@@ -72,6 +72,7 @@ object ValidColumnList {
  * @param copyColumnList Optional list of columns to use in copy statement.
  * @param sessionId Unqiue identifier for this write operation.
  * @param failedRowPercentTolerance Value between 0 and 1, percent of rows that are allowed to fail before the operation fails.
+ * @param mergeKey Key that two data sets are joined on in order to execute a merge statement
  */
 final case class DistributedFilesystemWriteConfig(jdbcConfig: JDBCConfig,
                                                   fileStoreConfig: FileStoreConfig,
@@ -82,7 +83,8 @@ final case class DistributedFilesystemWriteConfig(jdbcConfig: JDBCConfig,
                                                   copyColumnList: Option[ValidColumnList],
                                                   sessionId: String,
                                                   failedRowPercentTolerance: Float,
-                                                  filePermissions: ValidFilePermissions
+                                                  filePermissions: ValidFilePermissions,
+                                                  mergeKey: Option[String]
                                                  ) extends WriteConfig {
   private var overwrite: Boolean = false
 
--- a/connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
+++ b/connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
@@ -45,6 +45,7 @@ class VerticaDistributedFilesystemWritePipe(val config: DistributedFilesystemWri
           extends VerticaPipeInterface with VerticaPipeWriteInterface {
 
   private val logger = LogProvider.getLogger(classOf[VerticaDistributedFilesystemWritePipe])
+  private val tempTableName = new TableName(config.tablename.name + "_" + config.sessionId, None)
 
   /**
    * No write metadata required for configuration as of yet.
@@ -105,7 +106,7 @@ class VerticaDistributedFilesystemWritePipe(val config: DistributedFilesystemWri
       // Overwrite safety check
       _ <- if (config.isOverwrite && tableExistsPre) Left(DropTableError()) else Right(())
 
-      // Check if a view exists or temp table exits by this name
+      // Check if a view exists or temp table exists by this name
       viewExists <- tableUtils.viewExists(config.tablename)
       _ <- if (viewExists) Left(ViewExistsError()) else Right(())
       tempTableExists <- tableUtils.tempTableExists(config.tablename)
@@ -116,6 +117,9 @@ class VerticaDistributedFilesystemWritePipe(val config: DistributedFilesystemWri
       // Confirm table was created. This should only be false if the user specified an invalid target_table_sql
       tableExistsPost <- tableUtils.tableExists(config.tablename)
       _ <- if (tableExistsPost) Right(()) else Left(CreateTableError(None))
+
+      // Check if merge key was passed in. If so, this creates a temporary table, which will be merged into the target table.
+      _ <- if (config.mergeKey.isDefined) tableUtils.createTable(tempTableName, None, config.schema, config.strlen) else Right(())
 
       // Create the directory to export files to
       perm = config.filePermissions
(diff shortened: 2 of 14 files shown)


alexr-bq (changes requested):
Looks good, prototype so don't merge, create full ticket.

Need:
- list of keys
- true temporary table created as part of session
- unit tests
- integration tests
- consider interaction with overwrite mode
- consider interaction with copy_column_list

connector/src/main/scala/com/vertica/spark/config/WriteConfig.scala
```diff
@@ -72,6 +72,7 @@ object ValidColumnList {
  * @param copyColumnList Optional list of columns to use in copy statement.
  * @param sessionId Unqiue identifier for this write operation.
  * @param failedRowPercentTolerance Value between 0 and 1, percent of rows that are allowed to fail before the operation fails.
+ * @param mergeKey Key that two data sets are joined on in order to execute a merge statement
```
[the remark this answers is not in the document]
ravjotbrar: How about mergeConditions?

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
```diff
@@ -256,43 +285,53 @@ class VerticaDistributedFilesystemWritePipe(val config: DistributedFilesystemWri
...
       sessionId = config.sessionId
+
       rejectsTableName = "\"" +
         EscapeUtils.sqlEscape(tableName.substring(0,Math.min(tableNameMaxLength,tableName.length))) +
         "_" +
         sessionId +
         "_COMMITS" +
         "\""
 
-      copyStatement = buildCopyStatement(config.tablename.getFullTableName,
```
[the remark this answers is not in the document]
ravjotbrar: Good idea

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
```diff
@@ -147,6 +151,32 @@ class VerticaDistributedFilesystemWritePipe(val config: DistributedFilesystemWri
     s"COPY $targetTable $columnList FROM '$url' ON ANY NODE $fileFormat REJECTED DATA AS TABLE $rejectsTableName NO COMMIT"
   }
 
+  def buildMergeStatement(targetTableName: TableName, columnList: String, tempTable: String, mergeKey: Option[String]): String = {
+    var targetTable = targetTableName.getFullTableName
+    val updateColValues = schemaTools.getUpdateValues(jdbcLayer, targetTableName)
+    val insertColValues = schemaTools.getInsertValues(jdbcLayer, targetTableName)
+    val mergeKeyString = mergeKey match {
+      case Some(key) => key
+      case None => None
```
jonathanl-bq, line 160: How does this work? I'm assuming key is a String, so I'm surprised you can return a None here, since that's not of type String.

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
```diff
@@ -256,43 +285,53 @@ class VerticaDistributedFilesystemWritePipe(val config: DistributedFilesystemWri
...
+    val tableNameMaxLength = 30
     // Create url string, escape any ' characters as those surround the url
     val url: String = EscapeUtils.sqlEscape(s"${config.fileStoreConfig.address.stripSuffix("/")}/$globPattern")
-
-    val tableNameMaxLength = 30
-
+    val mergeKey= config.mergeKey match{
+      case Some(mergeKey) => true
+      case None => false
+    }
```
jonathanl-bq, line 294: I think you could use isDefined instead

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
```diff
@@ -256,43 +285,53 @@ class VerticaDistributedFilesystemWritePipe(val config: DistributedFilesystemWri
...
+
+      copyStatement = buildCopyStatement(fullTableName.toString(),
         columnList,
         url,
         rejectsTableName,
         "parquet"
       )
 
-      rowsCopied <- performCopy(copyStatement, config.tablename).left.map(_.context("commit: Failed to copy rows"))
+      rowsCopied <- if (config.mergeKey.isDefined) Right(performCopy(copyStatement, tempTableName).left.map(_.context("commit: Failed to copy rows"))) else Right(performCopy(copyStatement, config.tablename).left.map(_.context("commit: Failed to copy rows")))
```
jonathanl-bq, line 323: Split across multiple lines. Also, you should add different context for each case so we have a bit more debugging information.

connector/src/main/scala/com/vertica/spark/util/schema/SchemaTools.scala
```diff
@@ -320,5 +336,44 @@ class SchemaTools extends SchemaToolsInterface {
...
+    columnDefs.foreach(v => {
+      if (first) {
+        updateColValues = v.label + "=" + "temp." + v.label
+        first = false
+      }
+      else {
+        updateColValues += ", " + v.label + "=" + "temp." + v.label
+      }
+    })
+    updateColValues
```
jonathanl-bq, line 376: This is overly complicated. Just use mkString instead with a "," separator. See https://www.baeldung.com/scala/join-string-collection

I expect that this should be done without needing a foreach or mutable variables.

connector/src/main/scala/com/vertica/spark/util/schema/SchemaTools.scala
```diff
@@ -320,5 +336,44 @@ class SchemaTools extends SchemaToolsInterface {
...
+    // Create string of insert values to append to INSERT
+    columnDefs.foreach(v => {
+      if (first) {
+        tempColValues = "temp." + v.label
+        first = false
+      }
+      else {
+        tempColValues += ", temp." + v.label
+      }
+    })
```
jonathanl-bq, line 355: Same as below. Use mkString instead.

examples/basic-write/build.sbt
```diff
@@ -12,7 +12,7 @@
 // limitations under the License.
 
 scalaVersion := "2.12.12"
-name := "spark-vertica-connector-basic-write-example"
```
jonathanl-bq, line 15: Why this change? If we're changing this, we should make the other examples consistent with this too.

examples/run-example.sh
```diff
@@ -1,11 +1,11 @@
 if [ "$1" == "clean" ]
   then
-    wget -P ../../ https://apache.mirror.colo-serv.net/spark/spark-3.0.2/spark-3.0.2-bin-without-hadoop.tgz
+    wget -P ../../ https://mirror.dsrg.utoronto.ca/apache/spark/spark-3.1.2/spark-3.1.2-bin-without-hadoop.tgz
```
jonathanl-bq, line 3: Separate PR for these version changes

jonathanl-bq (changes requested):

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
ravjotbrar: I'll change it to "" for now, but will probably need to change it entirely once I update the string to a list.

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
ravjotbrar: I may get rid of this part of the code entirely, since it's not necessary.

examples/basic-write/build.sbt
ravjotbrar: Changed to basic-write because when we pass in the jar as an argument to run-example.sh, we instruct the user to use the directory name: `./run-example.sh [REPLACE WITH EXAMPLE DIR]/target/scala-2.12/spark-vertica-connector-[REPLACE WITH EXAMPLE DIR]-assembly-2.0.jar`. All other build.sbt files are consistent with this.

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
ravjotbrar: Good point.

connector/src/main/scala/com/vertica/spark/util/schema/SchemaTools.scala
ravjotbrar: Will this work even though columnDefs isn't a collection of strings?

jonathanl-bq: Call .map first

Issue #115, opened 2021-06-18
Use a smart client_label on each connection to vertica


NerdLogic:
Each connection made to Vertica should [set a client label](https://www.vertica.com/docs/10.1.x/HTML/Content/Authoring/ConnectingToVertica/ClientDriverMisc/SettingAClientConnectionLabel.htm) that identifies its purpose.  This helps identify and track down problems.  It also has the added benefit of helping quantify customer usage if customers share their scrutinize data with Vertica.

Here are some thoughts on what that could look like.  Some of these might not be possible, and we could add more later if needed.

vkspark-<cluster>[-config details][--purpose]
where <cluster> is some scrubbed string that identifies this spark cluster
where the optional [-config details] details might be:
-vs<version> - the version of the vertica spark connector
-sp<version> - the version of spark
-sc<version> - the version of scala
-py<version> - the version of python
-ps<version> - the version of pyspark
-n<node> - the spark node number?
-m<max node> - the number of spark nodes in cluster
-k - kerberos was used to connect
-p - password was used to connect
-c<conn number> - increment for each connection made from this node

e.g.
vspark-MyCluster5-vs2.0.1-sp3.1.0-sc2.13.1-n3-m12-k-c1--main
vspark-MyCluster5-vs2.0.1-sp3.1.0-sc2.13.1-n3-m12-k-c2--background_monitoring

PR #131, opened 2021-06-30
Add smart client label for JDBC connection

jonathanl-bq:
### Summary

This change adds a smart client label to expose more helpful debugging information.

### Description

This adds a smart client label. The label includes the version of the connector, the Spark version, the authentication method used, and a UUID for the connection.

### Related Issue
#115 

### Additional Reviewers
@alexr-bq 
@ravjotbrar

### DIFF
--- a/connector/src/main/scala/com/vertica/spark/datasource/jdbc/VerticaJdbcLayer.scala
+++ b/connector/src/main/scala/com/vertica/spark/datasource/jdbc/VerticaJdbcLayer.scala
@@ -295,6 +295,9 @@ class VerticaJdbcLayer(cfg: JDBCConfig) extends JdbcLayerInterface {
 
   def configureSession(fileStoreLayer: FileStoreLayerInterface): ConnectorResult[Unit] = {
     for {
+      connection <- this.connection
+      label <- Try{connection.getClientInfo("APPLICATIONNAME")}.toEither.left.map(e => handleJDBCException(e))
+      _ <- this.query(s"SELECT SET_CLIENT_LABEL('$label');")
       _ <- this.configureKerberosToFilestore(fileStoreLayer)
       _ <- this.configureAWSParameters(fileStoreLayer)
     } yield ()
--- a/connector/src/main/scala/com/vertica/spark/datasource/jdbc/VerticaJdbcLayer.scala
+++ b/connector/src/main/scala/com/vertica/spark/datasource/jdbc/VerticaJdbcLayer.scala
@@ -24,6 +24,7 @@ import com.vertica.spark.config.{BasicJdbcAuth, JDBCConfig, KerberosAuth, LogPro
 import com.vertica.spark.datasource.fs.FileStoreLayerInterface
 import com.vertica.spark.util.error.ErrorHandling.ConnectorResult
 import com.vertica.spark.util.general.Utils
+import hello.BuildInfo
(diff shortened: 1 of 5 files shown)


alexr-bq (approved):

connector/src/main/scala/com/vertica/spark/datasource/jdbc/VerticaJdbcLayer.scala
```diff
@@ -126,12 +127,22 @@ class VerticaJdbcLayer(cfg: JDBCConfig) extends JdbcLayerInterface {
...
         }, handleConnectionException).left.map(_.context("Initial connection was not valid.")))
   }
 
+  private def createClientLabel: String = {
+    val authMethod = this.cfg.auth match {
+      case BasicJdbcAuth(_, _) => "-p"
+      case KerberosAuth(_, _, _, _) => "-k"
+    }
+    val sparkVersion = SparkSession.active.sparkContext.version
+    "vspark" + "-vs" + BuildInfo.version + authMethod + "-sp" + sparkVersion + "-" + java.util.UUID.randomUUID.toString
```
NerdLogic, line 143: If you're going to include a UUID, you should use the same protocol and proceed it with a token that identifies what it is like "-uu".  BUT, I don't think there is any value in a UUID since the session id (select client_label, session_id from current_session) is already essentially unique for this database.  The session id is also in scrutin data, so across many databases we could join with other tables to make it unique if needed.

NerdLogic: A simple count (-cXXX) would make it easier to visually identify unique (mostly) connections.  Does spark have a way of keeping track of a sequence or something?  If not, process id might be good enough.

functional-tests/src/main/scala/com/vertica/spark/functests/JDBCTests.scala
```diff
@@ -190,5 +199,16 @@ class JDBCTests(val jdbcCfg: JDBCConfig) extends AnyFlatSpec with BeforeAndAfter
...
   }
 
+  it should "get the client label" in {
+    val result = jdbcLayer.query("SELECT GET_CLIENT_LABEL();")
 
+    result match {
+      case Right(rs) =>
+        assert(rs.next())
+        val label = rs.getString(1)
+        assert(label.contains("vspark-vs2.0.0-p-sp" + SparkSession.active.sparkContext.version))
```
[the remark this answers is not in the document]
jonathanl-bq: I'm probably only going to be able to verify that the version matches the regex. The BuildInfo.version comes from the connector project and I doubt there's an easy way to get that from our test project.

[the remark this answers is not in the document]
jonathanl-bq: Nevermind, it turns out this works.

connector/src/main/scala/com/vertica/spark/datasource/jdbc/VerticaJdbcLayer.scala
jonathanl-bq: Talked about this offline. Spark doesn't expose any sequence, and process id is not needed.

NerdLogic (approved):

PR #150, opened 2021-07-16
External table feature

alexr-bq:
### Summary

Provides an option to create an external table rather than uploading data in intermediary to Vertica.

### Description

- Implements new option for create_external_table
- Unit tests
- Integration tests

Will error out if this and other new feature, merge keys, are specified at the same time, as one always wants an existing table, and the other always wants to create a new table.

### Additional Reviewers

@ravjotbrar 
@jonathanl-bq


jonathanl-bq (approved):

ravjotbrar:
Just need to fix issue causing integration tests to fail

PR #159, opened 2021-07-23
Fix sbt deprecation warnings

ches:
### Summary

Since v1.5 sbt is becoming insistent about use of the ["unified slash syntax"](https://www.scala-sbt.org/1.x/docs/Migrating-from-sbt-013x.html#Migrating+to+slash+syntax) and issuing warnings on the good ol' `in` syntax.

In the interest of not being greeted with warnings immediately when you start to work on the project, let's make those go away.

### Description

Convert all scoped settings to use slash syntax.


jonathanl-bq (approved):
Approved. Thanks for the contribution!

alexr-bq (approved):

ravjotbrar (approved):

PR #188, opened 2021-07-29
Fix issue with duplicate IDs causing "File not found" error

alexr-bq:
### Summary

Fix duplicate ID issue causing read to fail under specific circumstances.

### Description

Previously, we would copy with a unique ID unless the filters were the same. This was based on a confusion between the planInputPartitions and scan build functions. The former is called twice per operation, and it was assumed the later was as well. This is not the case, so there is no need to filter the copying, we can always copy with a new unique ID.

### Additional Reviewers

@jonathanl-bq 
@ravjotbrar

### DIFF
--- a/connector/src/test/scala/com/vertica/spark/datasource/v2/VerticaV2SourceTest.scala
+++ b/connector/src/test/scala/com/vertica/spark/datasource/v2/VerticaV2SourceTest.scala
@@ -497,7 +497,6 @@ class VerticaV2SourceTests extends AnyFlatSpec with BeforeAndAfterAll with MockF
 
     // Implementation currently based on config, not identifier
     assert(catalog.tableExists(mock[Identifier]))
-
   }
 
   it should "catalog loads table on load or create" in {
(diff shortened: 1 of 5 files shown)


ravjotbrar (approved):

functional-tests/src/main/scala/com/vertica/spark/functests/EndToEndTests.scala
```diff
@@ -213,6 +213,7 @@ class EndToEndTests(readOpts: Map[String, String], writeOpts: Map[String, String
     TestUtils.dropTable(conn, tableName1)
   }
 
+
```
alexey-temnikov, line 216: nit: extra line

connector/src/test/scala/com/vertica/spark/datasource/v2/VerticaV2SourceTest.scala
```diff
@@ -497,6 +497,7 @@ class VerticaV2SourceTests extends AnyFlatSpec with BeforeAndAfterAll with MockF
 
     // Implementation currently based on config, not identifier
     assert(catalog.tableExists(mock[Identifier]))
+
```
jonathanl-bq, line 500: nit: Probably should remove this. I'm curious to know what scalafmt does about this though.

jonathanl-bq (approved):

Issue #154, opened 2021-07-21
Refactor examples' script code 


ravjotbrar:
There is common code between all the scripts for each example that should be extracted out to minimize redundancy. A possible solution is to move all of this code into the Dockerfile so that it is only run when the containers are composed up.

PR #213, opened 2021-08-10
Refactor run-example script

ravjotbrar:
### Summary

This change extracts out common code from different example scripts into the Dockerfile. 

### Description

Moved the download of Spark/Hadoop into the Dockerfile. This minimizes redundancy and ensures new examples that require their own scripts do not have to tweak the entire run-example script. This change also ensures that Spark and Hadoop are only downloaded when the docker containers are composed up for the first time.

### Related Issue

https://github.com/vertica/spark-connector/issues/154

### Additional Reviewers

@alexr-bq 
@jonathanl-bq

### DIFF
--- a/examples/pyspark/run-python-example.sh
+++ b/examples/pyspark/run-python-example.sh
@@ -1,48 +1,12 @@
-if [ "$1" == "clean" ]
-  then
-    yum install -y python3
-    wget -P ../../ https://archive.apache.org/dist/spark/spark-3.0.3/spark-3.0.3-bin-without-hadoop.tgz
-    wget -P ../../ https://archive.apache.org/dist/hadoop/common/hadoop-3.3.0/hadoop-3.3.0.tar.gz
-    cd ../../
-    tar xvf spark-3.0.2-bin-without-hadoop.tgz
-    tar xvf hadoop-3.3.0.tar.gz
-    mv spark-3.0.2-bin-without-hadoop/ /opt/spark
-    cd /opt/spark/conf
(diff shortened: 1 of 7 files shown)


examples/pyspark/run-python-example.sh
```diff
@@ -1,48 +1,12 @@
...
-    export PATH=$PATH:$SPARK_HOME/bin:$SPARK_HOME/sbin
-    start-master.sh
-    start-slave.sh spark://localhost:7077
-    cd ../../../spark-connector/examples
-    spark-submit --jars ../../connector/target/scala-2.12/spark-vertica-connector-assembly-2.0.1.jar sparkapp.py
+yum install -y python3
+export JAVA_HOME=/usr/lib/jvm/jre-11-openjdk
+export SPARK_DIST_CLASSPATH=$(/hadoop-3.3.0/bin/hadoop classpath)
+export SPARK_HOME=/opt/spark
+export PATH=$PATH:$SPARK_HOME/bin:$SPARK_HOME/sbin
```
jonathanl-bq, line 5: These environment variables should also be set in the Dockerfile. You can use "ENV <command>" to do this.

jonathanl-bq (changes requested):

examples/pyspark/run-python-example.sh
jonathanl-bq: Discussed offline and SPARK_DIST_CLASSPATH is problematic. Until we come up with a way to solve that problem, we're just going to have to leave it.

jonathanl-bq (approved):

Issue #225, opened 2021-08-30
Update new connector version to reflect the 2.0.2 release


ravjotbrar:
Update connector version in our repo where it was originally 2.0.1 to 2.0.0

PR #226, opened 2021-08-30
Update connector version to 2.0.2

ravjotbrar:
### Summary

Update connector version to 2.0.2 to reflect our new release

### Description

Replaced 2.0.1 with 2.0.2 in our project

### Related Issue

https://github.com/vertica/spark-connector/issues/225

### Additional Reviewers

@alexr-bq 
@jonathanl-bq

### DIFF
--- a/examples/demo/build.sbt
+++ b/examples/demo/build.sbt
@@ -21,5 +21,5 @@ resolvers += "jitpack" at "https://jitpack.io"
 
 libraryDependencies ++= Seq(
   "com.typesafe" % "config" % "1.4.1",
-  "com.vertica.spark" % "vertica-spark" % "2.0.2-all"
+  "com.vertica.spark" % "vertica-spark" % "2.0.2-slim"
 )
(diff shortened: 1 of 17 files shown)


jonathanl-bq (approved):

examples/demo/build.sbt
```diff
@@ -14,12 +14,12 @@
...
-version := "2.0.1"
+version := "2.0.2"
 
 resolvers += "Artima Maven Repository" at "https://repo.artima.com/releases"
 resolvers += "jitpack" at "https://jitpack.io"
 
 libraryDependencies ++= Seq(
   "com.typesafe" % "config" % "1.4.1",
-  "com.vertica.spark" % "vertica-spark" % "2.0.1-0"
+  "com.vertica.spark" % "vertica-spark" % "2.0.2-0"
```
alexey-temnikov, line 24: Seems like we should use `2.0.2-slim` instead, as `2.0.2-0` is not published.

alexey-temnikov (changes requested):

examples/demo/build.sbt
ravjotbrar: I thought if we don't specify either, then -all is the default?

alexey-temnikov (approved):

PR #231, opened 2021-09-02
Extra Docker Automation

alexey-temnikov:


docker/sandbox-clientenv.sh
```diff
@@ -8,6 +8,10 @@ function configure_db() {
 }
 
 function configure_hdfs() {
+  docker exec hdfs service ssh start
+  docker exec hdfs start-dfs.sh
+  sleep 10
```
jonathanl-bq, line 13: I guess the sleep is necessary, but I'm still curious to know why it's here.

docker/sandbox-clientenv.sh
```diff
@@ -8,6 +8,10 @@ function configure_db() {
...
+#  docker exec hdfs hdfs dfsadmin -safemode forceExit
```
jonathanl-bq, line 14: Do we need this?

alexey-temnikov: No, needs to be removed.

docker/sandbox-clientenv.sh
alexey-temnikov: Artifact from testing, suggesting to remove and confirm no impact.

Issue #200, opened 2021-08-04
External Tables support for existing data


jonathanl-bq:
Feature request: use the external table capabilities, and the other connectivity support in the connector such as Kerberos support, to provide a method for creating an external table based on existing parquet data already on disk.

![ExternalTables](https://user-images.githubusercontent.com/7035336/131002421-f354d761-297e-4d65-86f3-e86875ad654d.png)

PR #233, opened 2021-09-13
External tables for Existing Data

ravjotbrar:
### Summary

This change implements the second use-case for external tables. Users will now be able to create an external table out of existing parquet data on disk. This is done by using the spark connector to write an _empty_ DF and supplying the path to the existing data in the staging_fs_url config option. In addition, the user must specify the create_external_table config option as "existing-data" if they want to use existing data or "new-data" if they want to create an external table out of spark data. 

### Description

Changed the create_external_table option from Bool to Option[CreateExternalOption]; the option now accepts "existing-data", "new-data", or None. If the user passes in the "existing-data", the write operation to HDFS is bypassed and a method to infer the schema of the existing data is called. If a user passes in a non-empty dataframe while trying to create an external table using existing data, the connector will error out.

### Related Issue
https://github.com/vertica/spark-connector/issues/200

### Additional Reviewers

@jonathanl-bq 
@alexr-bq 
@valentina-bq 
@alexey-temnikov


README.md
```diff
@@ -122,8 +122,9 @@ Below is a detailed list of connector options that are used in the options map:
...
+| create_external_table                          | String                                              | An option to create an external table, either out of existing parquet data in HDFS or by writing data from Spark to HDFS. For the latter use-case, specify this option as "new-data". For the former use-case, specify this option as "existing-data". If you are creating an external table from existing parquet data, please ensure you specify the path to the directory containing the appropriate parquet files in the staging_fs_url option ("webhdfs://hdfs:50070/data/data.parquet/"). Also, ensure that the spark dataframe you are writing is empty.                                                                                  | No
```
alexr-bq, line 127: I think we should maintain support for "true" here as an option for anyone using the current version.

ravjotbrar: Done.

connector/src/main/scala/com/vertica/spark/datasource/core/DSConfigSetup.scala
```diff
@@ -113,15 +121,16 @@ object DSConfigSetupUtils {
...
   }
 
-  def getCreateExternalTable(config: Map[String, String]): ValidationResult[Boolean] = {
+  def getCreateExternalTable(config: Map[String, String]): ValidationResult[Option[CreateExternalTableOption]] = {
     config.get("create_external_table") match {
       case Some(str) =>
         str match {
-          case "true" => true.validNec
-          case "false" => false.validNec
+          case "new-data" => Some(NewData).validNec
```
jonathanl-bq, line 128: nit: I believe you can use an OR pattern here to catch the "true" case in here as well to do this in 1 line.

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
```diff
@@ -142,16 +143,28 @@ class VerticaDistributedFilesystemWritePipe(val config: DistributedFilesystemWri
...
-
     fileStoreLayer.openWriteParquetFile(filename)
+
   }
 
   def writeData(data: DataBlock): ConnectorResult[Unit] = {
-    fileStoreLayer.writeDataToParquetFile(data)
+    config.createExternalTable match {
+      case Some(value) =>
+        if(value.toString == "existing-data") {
```
jonathanl-bq, line 153: Why do we check the string representation? The value is an ADT, so we could just pattern match here instead.

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
```diff
@@ -142,16 +143,28 @@ class VerticaDistributedFilesystemWritePipe(val config: DistributedFilesystemWri
...
+          fileStoreLayer.writeDataToParquetFile(data)
+        }
+      case None => fileStoreLayer.writeDataToParquetFile(data)
+    }
   }
 
   def endPartitionWrite(): ConnectorResult[Unit] = {
-    fileStoreLayer.closeWriteParquetFile()
+    config.createExternalTable match {
+      case Some(value) => if(value.toString == "existing-data") Right(()) else fileStoreLayer.closeWriteParquetFile()
```
jonathanl-bq, line 165: same here, should not use toString.

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
```diff
@@ -372,13 +412,20 @@ class VerticaDistributedFilesystemWritePipe(val config: DistributedFilesystemWri
...
+
       _ <- tableUtils.createExternalTable(
-        tablename = config.tablename,
-        targetTableSql = config.targetTableSql,
-        schema = config.schema,
-        strlen = config.strlen,
-        urlToCopyFrom = url
-      )
+                tablename = config.tablename,
+                if(createExternalTableOption == "existing-data") Some(createExternalTableStmt.toString) else config.targetTableSql,
```
jonathanl-bq, line 424: same here, avoid checking the string representation.

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
```diff
@@ -372,13 +412,20 @@ class VerticaDistributedFilesystemWritePipe(val config: DistributedFilesystemWri
 
     val ret = for {
       _ <- jdbcLayer.configureSession(fileStoreLayer)
+
+      createExternalTableOption = config.createExternalTable match {
```
jonathanl-bq, line 416: The typing here is odd. One case returns None, which is an Option, but the other case returns a String. The only common supertype for these two types would be Any, which we really should avoid unless absolutely necessary.

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
```diff
@@ -372,13 +412,20 @@ class VerticaDistributedFilesystemWritePipe(val config: DistributedFilesystemWri
...
+        case Some(value) => value.toString
+        case None => None
+      }
+      createExternalTableStmt <- if(createExternalTableOption == "existing-data") inferExternalTableSchema else Right(())
```
jonathanl-bq, line 420: Avoid checking the string here.

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
```diff
@@ -400,9 +447,19 @@ class VerticaDistributedFilesystemWritePipe(val config: DistributedFilesystemWri
     val globPattern: String = "*.parquet"
 
     // Create url string, escape any ' characters as those surround the url
-    val url: String = EscapeUtils.sqlEscape(s"${config.fileStoreConfig.address.stripSuffix("/")}/$globPattern")
+    val url: String = config.createExternalTable match {
+      case Some(value) =>
+        if(value.toString == "new-data") {
```
jonathanl-bq, line 452: Remove String check

connector/src/main/scala/com/vertica/spark/datasource/v2/VerticaDatasourceV2Write.scala
```diff
@@ -136,11 +134,26 @@ class VerticaBatchWriter(config: WriteConfig, writer: DSWriterInterface) extends
...
   // Construct a unique identifier string based on the partition and task IDs we've been passed for this operation
 
-  writer.openWrite() match {
-    case Right(_) => ()
-    case Left(err) => ErrorHandling.logAndThrowError(logger, err)
+  config match {
+    case cfg: DistributedFilesystemWriteConfig =>
+      cfg.createExternalTable match {
+        case Some(value) =>
+          if (value.toString != "existing-data") {
```
jonathanl-bq, line 141: Remove String check.

examples/basic-write-example/src/main/resources/application.conf
```diff
@@ -6,4 +6,3 @@ functional-tests {
   password=""
   filepath="webhdfs://hdfs:50070/data/"
 }
-
```
jonathanl-bq, line 9: Please leave this extra line at the end of the file.

jonathanl-bq (changes requested):
Please also remove and add to .gitignore the hdfs.keytab, vertica.keytab, and hdfs.cert files.

examples/external-table-example/build.sbt
```diff
@@ -23,7 +23,7 @@ libraryDependencies ++= Seq(
   "com.typesafe" % "config" % "1.4.1",
   "org.apache.spark" %% "spark-core" % "3.0.0",
   "org.apache.spark" %% "spark-sql" % "3.0.0",
-  "com.vertica.spark" % "vertica-spark" % "2.0.2-slim"
+  //"com.vertica.spark" % "vertica-spark" % "2.0.2-slim"
```
jonathanl-bq, line 26: I'd rather just remove this and maybe leave a note to update this later. Don't forget to update the README for this example as well with instructions on building the connector JAR and placing it in the lib/ folder. We should always make sure our documentation reflects the current state of the project, even if we are going to update it again next release.

connector/src/test/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipeTest.scala
```diff
@@ -600,9 +609,34 @@ class VerticaDistributedFilesystemWritePipeTest extends AnyFlatSpec with BeforeA
...
+
+    val jdbcLayerInterface = mock[JdbcLayerInterface]
+    (jdbcLayerInterface.query _).expects("SELECT INFER_EXTERNAL_TABLE_DDL(\'hdfs://example-hdfs:8020/tmp/testtable.parquet/*.parquet\',\'testtable\')",*).returning(Right(getStringResultSet))
+    (jdbcLayerInterface.commit _).expects().returning(Right(()))
+    (jdbcLayerInterface.close _).expects().returning(Right(()))
+    (jdbcLayerInterface.configureSession _).expects(fileStoreLayerInterface).returning(Right(()))
+
+    val schemaToolsInterface = mock[SchemaToolsInterface]
+    val tableUtils = mock[TableUtilsInterface]
+    (tableUtils.createExternalTable _).expects(tname, Some("create external table \"testtable\"(\"col1\" int) as copy from \'hdfs://example-hdfs:8020/tmp/testtable.parquet/*.parquet\' parquet"), config.schema, config.strlen, "hdfs://example-hdfs:8020/tmp/testtable.parquet/*.parquet").returning(Right(()))
```
jonathanl-bq, line 627: Watch the line length here. I think this is a bit too long and should be split across multiple lines. Ideally, I'd like to avoid having to scroll horizontally just to review this file.

connector/src/main/scala/com/vertica/spark/util/schema/SchemaTools.scala
```diff
@@ -425,5 +434,37 @@ class SchemaTools extends SchemaToolsInterface {
...
+        val colName = col.substring(indexOfFirstDoubleQuote, indexOfSpace)
+        val fieldType = schema.collect {
+          case field if(addDoubleQuotes(field.name) == colName) => field.dataType.simpleString
+        }
+        if(fieldType.nonEmpty) {
+          colName + " " + fieldType.head
+        }
+        else {
+          Left(UnknownColumnTypesError)
+          col
```
jonathanl-bq, line 459: This makes no sense. The Left value here is going to be swallowed. What are you trying to do here?

connector/src/main/scala/com/vertica/spark/util/schema/SchemaTools.scala
```diff
@@ -425,5 +434,37 @@ class SchemaTools extends SchemaToolsInterface {
...
+
+    val schemaList = schemaString.split(",").toList
+
+    val updatedSchema: String = schemaList.map(col => {
+      if(col.contains("UNKNOWN")) {
+        val indexOfFirstDoubleQuote = col.indexOf("\"")
+        val indexOfSpace = col.indexOf(" ", indexOfFirstDoubleQuote)
+        val colName = col.substring(indexOfFirstDoubleQuote, indexOfSpace)
+        val fieldType = schema.collect {
+          case field if(addDoubleQuotes(field.name) == colName) => field.dataType.simpleString
```
jonathanl-bq, line 452: This is non-total. There should have been a compiler warning about this, but you aren't handling the other case where the if condition fails.

jonathanl-bq (changes requested):

README.md
alexr-bq: To clarify, I think we should also accept 'true' going forward in code so not break compatibility.

connector/src/main/scala/com/vertica/spark/util/schema/SchemaTools.scala
jonathanl-bq: Spoke to Ravjot about this. Apparently, it's supposed to be a partial function, since that's what the Spark API wants.

jonathanl-bq (approved):

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
```diff
@@ -102,10 +103,10 @@ class VerticaDistributedFilesystemWritePipe(val config: DistributedFilesystemWri
       _ = setSparkCalendarConf()
 
       // If overwrite mode, remove table and force creation of new one before writing
-      _ <- if(config.isOverwrite && config.mergeKey.isEmpty) tableUtils.dropTable(config.tablename) else Right(())
+      _ <- if (config.isOverwrite && config.mergeKey.isEmpty) tableUtils.dropTable(config.tablename) else Right(())
 
       // Creating external table and merging at the same time not supported
-      _ <- if (config.createExternalTable && config.mergeKey.isDefined) Left(CreateExternalTableMergeKey()) else Right(())
+      _ <- if ((config.createExternalTable.isDefined) && config.mergeKey.isDefined) Left(CreateExternalTableMergeKey()) else Right(())
```
jonathanl-bq, line 109: I don't think these parentheses around config.createExternalTable.isDefined are needed.

jonathanl-bq (approved):

Issue #236, opened 2021-09-15
Simplify param to run-example script


ravjotbrar:
At the moment, we need to pass the entire path to the assembled jar file of the example we want to run. Will attempt to simplify this to just the example name as the rest of the path is the same for all examples.

PR #237, opened 2021-09-15
Simplify param passed into run-example

ravjotbrar:
### Summary

Before this change, we needed to pass the entire path to the assembled jar file of the example we want to run. Now, we only need to pass in the example's name as a parameter.

### Description

Same as above. Changes made in run-example.sh and readme. 

### Related Issue

https://github.com/vertica/spark-connector/issues/236

### Additional Reviewers

@alexr-bq 
@jonathanl-bq 
@valentina-bq

### DIFF
--- a/examples/kerberos-example/run-example.sh
+++ b/examples/kerberos-example/run-example.sh
@@ -4,4 +4,4 @@ export SPARK_HOME=/opt/spark
 export PATH=$PATH:$SPARK_HOME/bin:$SPARK_HOME/sbin
 start-master.sh
 start-worker.sh spark://localhost:7077
-spark-submit --master spark://localhost:7077 --driver-java-options '-Djava.security.auth.login.config=/spark-connector/docker/client-krb/jaas.config' $1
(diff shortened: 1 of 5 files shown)


examples/kerberos-example/run-example.sh
```diff
@@ -4,4 +4,4 @@ export SPARK_HOME=/opt/spark
 export PATH=$PATH:$SPARK_HOME/bin:$SPARK_HOME/sbin
 start-master.sh
 start-worker.sh spark://localhost:7077
-spark-submit --master spark://localhost:7077 --driver-java-options '-Djava.security.auth.login.config=/spark-connector/docker/client-krb/jaas.config' $1
+spark-submit --master spark://localhost:7077 --driver-java-options '-Djava.security.auth.login.config=/spark-connector/docker/client-krb/jaas.config' ./target/scala-2.12/spark-vertica-connector-$1-assembly-2.0.2.jar
```
ravjotbrar, line 7: I just realized since this is in the kerberos-example dir, we don't need to pass a parameter. To make the run-example script more clear, we can change the name to run-kerberos-example.sh (similar to the pyspark and sparklyr examples)

jonathanl-bq (approved):

PR #240, opened 2021-09-16
Update README.md

NerdLogic:
### Summary

Minor changes to readme

### Description

fixed version of assembly jar
added text on building assembly and cloning

### Additional Reviewers

@jonathanl-bq

### DIFF
--- a/examples/kerberos-example/README.md
+++ b/examples/kerberos-example/README.md
@@ -14,7 +14,7 @@ Change directory to the example project:
 cd /spark-connector/examples/kerberos-example
 ```
 
-Build the assembly if it hasn't been built already
+Build the assembly if it hasn't been built already.  Note: building within docker may be slower.
 ```
 sbt assembly
 ```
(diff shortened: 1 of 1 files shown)


examples/kerberos-example/README.md
```diff
@@ -16,9 +14,14 @@ Change directory to the example project:
 cd /spark-connector/examples/kerberos-example
 ```
 
+Build the assembly if it hasn't been built already
```
jonathanl-bq, line 17: I think it may be worth mentioning that this runs significantly faster outside of the Docker container.

jonathanl-bq (approved):
Minor suggestion, but otherwise this looks good.

examples/kerberos-example/README.md
NerdLogic: Ok, I added a note.  I don't have sbt installed natively, so I prefer the lower admin costs of running sbt in docker.

Issue #242, opened 2021-09-17
Fix kerberos environment setup on Windows


ravjotbrar:
The batch file that creates the kerberized docker environment is not sourcing the IP address for hdfs correctly. At the moment it is incorrectly sourced as %undefined% in the sandbox-clientenv.bat file.

PR #244, opened 2021-09-18
Fixed kerberos environment setup on Windows.

valentina-bq:
### Summary
Fixed kerberos environment setup on Windows.

### Description
Moved kerberos environment setup into the separate script named `sandbox-kerberos-clientenv.bat`, which is called indirectly from the main `sandbox-clientenv.bat` when `kerberos` parameter is provided. This way user experience doesn't change, although running the new script directly would work the same.

### Related Issue
https://github.com/vertica/spark-connector/issues/242

### Additional Reviewers
@jonathanl-bq 
@alexey-temnikov 
@alexr-bq 
@ravjotbrar

### DIFF
--- a/docker/sandbox-clientenv.bat
+++ b/docker/sandbox-clientenv.bat
@@ -1,23 +1,7 @@
 @echo off
 
 IF "%~1" == "kerberos" (
-	echo "running kerberos docker compose"
-	docker compose -f docker-compose-kerberos.yml up -d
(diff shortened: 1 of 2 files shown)


jonathanl-bq:
This looks good so far, but there should be some documentation for Windows users to go along with this, as described in the ticket.

jonathanl-bq (approved):

PR #256, opened 2021-09-29
Update version to 2.0.3

ravjotbrar:
### Summary

Update version in the spark connector to reflect new release

### Description

Replaced every instance of 2.0.2 with 2.0.3

### Additional Reviewers

@alexr-bq 
@jonathanl-bq


docker/vertica-hdfs-config/hadoop-kerberized/hdfs.cert
```diff
@@ -1,21 +1,21 @@
 -----BEGIN CERTIFICATE-----
```
jonathanl-bq, line 1: this file should not be checked in

jonathanl-bq (approved):

Issue #259, opened 2021-10-13
Support file paths without the filesystem modifier at the front


alexr-bq:
If hdfs is configured, support a path such as 'user/v1/parquet' rather than 'hdfs://user/v1/parquet'. We should be able to pick this up based on the given hadoop configuration, and support hdfs / webhdfs / s3 paths here.

PR #280, opened 2021-11-27
Support hdfs uri without the filesystem modifier at the front

ravjotbrar:
### Summary

This change allows users to specify the filepath portion of the uri without the filesystem modifier at the front.

### Description

Filesystem modifer is now sourced from core-site.xml and prefixed with the supplied filepath. Most of the change is reflected in FileStoreConfig.

### Related Issue

https://github.com/vertica/spark-connector/issues/259

### Additional Reviewers

@jonathanl-bq 
@alexr-bq

### DIFF
--- a/connector/src/main/scala/com/vertica/spark/config/FileStoreConfig.scala
+++ b/connector/src/main/scala/com/vertica/spark/config/FileStoreConfig.scala
@@ -13,6 +13,12 @@
 
 package com.vertica.spark.config
 
+import org.apache.spark.sql.SparkSession
+import org.apache.hadoop.hdfs.client.HdfsClientConfigKeys
+import com.vertica.spark.util.error.{NoSparkSessionFound, HDFSConfigError}
+import com.vertica.spark.util.error.ErrorHandling.ConnectorResult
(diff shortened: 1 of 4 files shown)


connector/src/main/scala/com/vertica/spark/config/FileStoreConfig.scala
```diff
@@ -47,16 +53,39 @@ case class AWSOptions(
...
+    if(baseAddress.startsWith("/")) {
+      SparkSession.getActiveSession match {
+        case Some(session) =>
+          val hadoopConf = session.sparkContext.hadoopConfiguration
+          val filepath = Option(hadoopConf.get("fs.defaultFS")) match {
+            case Some(path) => Right(path)
+            case None => Left(HDFSConfigError())
+          }
+          filepath
+        case None => Left(NoSparkSessionFound())
```
jonathanl-bq, line 66: Might be nice to use .context() to add context on where this error is coming from as it propagates up, since you're reusing the error. That way we can easily differentiate which part of the codebase it's coming from.

jonathanl-bq (approved):
Approved with 1 suggestion.

Issue #251, opened 2021-09-27
Provide option for assigning full name of export location rather than using unique ID


alexr-bq:
Request: when using external tables, it could be a nicer user experience to specify the full location of exported parquet files to base the external table off of. This is so it is easier to then query the written parquet data. Currently, we always create a unique identifier that is used in the path. 

Provide an option to replace the (unique-id/tablename) path format with a user provided value.

PR #286, opened 2021-12-02
Remove unique identifier from external table export location

jeremyp-bq:
### Summary

Remove unique identifier from external table export location.

### Description

Previously when exporting data as an external table it would append a unique identifier to to the export path.  This change removes that identifier from the path so that it is easier to query it directly (since we already know the full path).

When setting `create_external_table` to `new-data` or `true` the data will be saved under `staging_fs_url`, without a unique identifier appended to the path.

For example, assuming `create_external_table` = `new-data` and `staging_fs_url` = `webhdfs://hdfs:50070/3.1.1/mytest`, here is where the data was previously stored and where it is stored with this change:
```
# BEFORE
webhdfs://hdfs:50070/3.1.1/mytest/d629c41f_4304_4529_884e_37a98912ef23/0-0.parquet

# AFTER
webhdfs://hdfs:50070/3.1.1/mytest/0-0.parquet
```

### Related Issue

#251

### Additional Reviewers

@alexey-temnikov
@alexr-bq
@jonathanl-bq
@ravjotbrar


connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
```diff
@@ -142,15 +142,16 @@ class VerticaDistributedFilesystemWritePipe(val config: DistributedFilesystemWri
...
+      _ <- if(existingData) Right(()) else fileStoreLayer.createDir(location, perm.toString)
 
       // Create job status table / entry
       _ <- tableUtils.createAndInitJobStatusTable(config.tablename, config.jdbcConfig.auth.user, config.sessionId, if(config.isOverwrite) "OVERWRITE" else "APPEND")
     } yield ()
   }
 
   def startPartitionWrite(uniqueId: String): ConnectorResult[Unit] = {
-    val address = config.fileStoreConfig.address
+    val address = if (config.createExternalTable.isDefined) config.fileStoreConfig.externalTableAddress else config.fileStoreConfig.address
```
alexr-bq, line 154: Since this same thing is checked a few times, could this check go into a function, here or in writeconfig?

jeremyp-bq: Done.

I created a function (`getAddress`) in this class since it seemed easier to read than trying to add one in `WriteConfig` (with its nested `FileStoreConfig`).

alexr-bq (approved):

ravjotbrar (approved):

Issue #283, opened 2021-12-02
Connector option to prevent cleanup


ravjotbrar:
Add a boolean connector option that defaults to false, but can be set to true if users want to keep parquet files in intermediary store after a read or write.

PR #287, opened 2021-12-03
Parquet cleanup

ravjotbrar:
### Summary

Created a connector option which prevents the cleanup of parquet files in the intermediary file store during both a read and a write. This option was created so users could debug data that was written or exported as parquet data. 

### Description

Added a new connector option called prevent_cleanup, which defaults to "false", but may be set to "true" if users intend to keep the parquet files after the spark job is complete. Added unit tests and functional tests for this new feature. 

### Related Issue

https://github.com/vertica/spark-connector/issues/283

### Additional Reviewers

@jeremyp-bq 
@alexr-bq 
@jonathanl-bq


connector/src/main/scala/com/vertica/spark/config/FileStoreConfig.scala
```diff
@@ -52,7 +52,7 @@ case class AWSOptions(
  * @param baseAddress Address to use in the intermediate filesystem
  * @param sessionId Unique id for a given connector operation
  */
-final case class FileStoreConfig(baseAddress: String, sessionId: String, awsOptions: AWSOptions) {
+final case class FileStoreConfig(baseAddress: String, preventCleanup: Boolean, sessionId: String, awsOptions: AWSOptions) {
```
jeremyp-bq, line 55: JavaDocs should to be updated.  May also consider moving this as the 3rd or 4th parameter as baseAddress and sessionId are used together.

examples/basic-write-example/build.sbt
```diff
@@ -14,7 +14,7 @@
 scalaVersion := "2.12.12"
 name := "spark-vertica-connector-basic-write-example"
 organization := "com.vertica"
-version := "3.0.0"
+version := "2.0.4"
```
jeremyp-bq, line 17: Should this be `3.0.0` like the other examples?

connector/src/main/scala/com/vertica/spark/config/FileStoreConfig.scala
ravjotbrar: Updated.

connector/src/main/scala/com/vertica/spark/datasource/core/DSConfigSetup.scala
```diff
@@ -389,6 +389,18 @@ object DSConfigSetupUtils {
     }
   }
 
+  def getPreventCleanup(config: Map[String, String]) : ValidationResult[Boolean] = {
```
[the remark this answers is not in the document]
ravjotbrar: I believe since we're sourcing config options here, we prepend all of the methods with "get"

examples/basic-write-example/build.sbt
ravjotbrar: Good catch. Updated.

jeremyp-bq (approved):

connector/src/main/scala/com/vertica/spark/util/error/ErrorHandling.scala
```diff
@@ -203,6 +203,10 @@ case class InvalidCreateExternalTableOption() extends ConnectorError {
   override def getFullContext: String = "The 'create_external_table' param is invalid. Please specify " +
     "'new-data' or 'existing-data'."
 }
+case class InvalidPreventCleanupOption() extends ConnectorError {
+  override def getFullContext: String = "The 'prevent_cleanup' param is invalid. Please specify " +
+    "'true' or 'false'. This option will default to false"
```
jonathanl-bq, line 208: This error message should be more specific. It says this option will default to false, but that almost implies that it defaults to false in the case where this error shows up, which is certainly not true. I think it would be better to elaborate on this and say it defaults to false when no value is specified.

ravjotbrar: Agreed. I believe I was in the process of completing the sentence but never did for some reason. Updated.

README.md
```diff
@@ -124,14 +124,15 @@ Below is a detailed list of connector options that are used in the options map:
...
+| prevent_cleanup                                | String                                              | An option to prevent cleanup in the intermediary filestore layer during a read or a write. The default value is "false", but this may be set to true to keep parquet files after a Spark job is ran.                                                                                                                                                                                                                                                                                                                                                                                                         | No                                             |                      "false"|
```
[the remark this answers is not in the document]
jonathanl-bq: Every parameter is a string though. Spark configs are literally a Map[String, String]. I think it's better to give users an idea of what the format of the string should be.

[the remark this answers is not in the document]
ravjotbrar: Agreed. Updated.

jonathanl-bq (approved):

Issue #284, opened 2021-12-02
Handle error where 0 rows are copied and 0 rows are rejected


ravjotbrar:
The error: Row count is above error tolerance threshold

- Likely happens because data is not written to hdfs correctly
- I have repro'd this issue once before when using hdfs uri instead of webhdfs in a kerberized environment

Issue #253, opened 2021-09-28
Fix some end-end tests error types


ravjotbrar:
Following are the tests in CI along with their expected errors:
```
should Fail DataFrame with Complex type array (CommitError)

should reject invalid rows (CommitError)

should Vertica column type mismatch in Append mode. (CommitError) 

should Should throw clear error message if data frame contains a complex data type not supported by Vertica. (CommitError)

should Should not try to save an empty dataframe. (CommitError)

should Fail if schema doesn't match table data type (TempTableExistsError) 

should Reject 1/5 of rows, and hence not pass failed_rows_percent_tolerance.  Append mode. (TempTableExistsError)

should fail to generate default copy by name and by position if cols names and count are different (CommitError)

should "fail to generate default copy by name and by position if cols names and count are different"  (CommitError)
```
The action item is to update these tests to an appropriate error type, whether it's the expected error above or a different error that describes the failure better. In addition, a couple of the tests have a typo where we are writing "should" twice. This should be updated.

ravjotbrar:
#1, #3, #4, #8 have been updated to their expected error types.

#2, #5 fail at tolerance error correctly

#6, #7 fail at commit error correctly

This issue will be closed along with #284

PR #291, opened 2021-12-08
Handle error when copy is invalid

ravjotbrar:
### Summary

Prior to this change, users would not get a meaningful error when attempting to perform a valid copy. Instead, the connector would give an error where 0 rows were copied and 0 were rejected. This change addresses that issue, along with updating error types in some of the end-end tests. 

### Description

In the commit step, the copy is matched to a Right(int) or Left(err) to indicate the number of rows copied or an error if the copy step could not be performed. As a result, some end to end tests required their error types to be changed from FaultToleranceTestFail to CommitError. In addition, updated some minor nits in the end-end tests. 

### Related Issue

https://github.com/vertica/spark-connector/issues/284
https://github.com/vertica/spark-connector/issues/253

### Additional Reviewers

@jeremyp-bq 
@alexey-temnikov 
@jonathanl-bq 
@alexr-bq

### DIFF
--- a/connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
+++ b/connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
@@ -430,16 +430,10 @@ class VerticaDistributedFilesystemWritePipe(val config: DistributedFilesystemWri
       _ = logger.info("The copy statement is: \n" + copyStatement)
 
       rowsCopied <- if (config.mergeKey.isDefined) {
-        performCopy(copyStatement, config.tablename) match{
-          case Right(numRows) => Right(numRows)
-          case Left(err) => Left(err)
-        }
+        performCopy(copyStatement, config.tablename)
       }
       else {
-        performCopy(copyStatement, config.tablename) match{
-          case Right(numRows) => Right(numRows)
-          case Left(err) => Left(err)
-        }
+        performCopy(copyStatement, config.tablename)
       }
 
       faultToleranceResults <- testFaultTolerance(rowsCopied, rejectsTableName)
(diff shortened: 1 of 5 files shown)


jeremyp-bq (approved):

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
```diff
@@ -430,13 +430,13 @@ class VerticaDistributedFilesystemWritePipe(val config: DistributedFilesystemWri
       _ = logger.info("The copy statement is: \n" + copyStatement)
 
       rowsCopied <- if (config.mergeKey.isDefined) {
-                      Right(performCopy(copyStatement, tempTableName).left.map(_.context("commit: Failed to copy rows into temp table")))
-                    }
-                    else {
-                      Right(performCopy(copyStatement, config.tablename).left.map(_.context("commit: Failed to copy rows into target table")))
-                    }
+        performCopy(copyStatement, config.tablename)
```
jonathanl-bq, line 433: This change makes no sense. The original code added context to each branch of this if statement to ensure that we could trace which path the code went down in the error. Now we do a performCopy in either case with no distinction. Furthermore, there is no difference between the two cases, making the entire config.mergeKey.isDefined check pointless.

jonathanl-bq (changes requested):

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
jonathanl-bq: If you don't think we need the context here, then I'm fine with removing it. However, there is still no distinction between the two cases. Abstractly, you have an if A then B else B, which is the same as B.

ravjotbrar: No if you take a look again, the distinction is in the destination of where the copy is performed (temptable for merge and config.tablename otherwise). I don't mind adding the context back in, as I think it can be helpful, but there is still a distinction between the two cases.

ravjotbrar: I see the issue now. The copy paste was a typo, will update.

jonathanl-bq (approved):

Issue #293, opened 2021-12-08
Provide configuration options for rejected data table


jeremyp-bq:
With the change done in #275 we now persist the rejected rows data to a table on write (whenever there is at least one rejected row a new table will be created per transaction).

This ticket is to provide the option to configure this functionality.  For example, we could add a `persist_rejected_data` option (defaults to `true`) and a `rejected_data_table` option (defaults to the current `<table>_<session_id>_COMMITS` or something more descriptive like `<table>_<session_id>_rejected_data`).

Issue #275, opened 2021-11-18
Look into how rejected rows are handled in v2 connector


ravjotbrar:
Modify an example to contain rejected rows and see if those rows are contained in rejects table. We expect rows with issues to be in this table.

jeremyp-bq:
Note that rows are only rejected during the copying/committing of data from HDFS to Vertica.  This means that the data must be valid in HDFS, but for whatever reason it does not satisfy the constraints in Vertica.

A good way to test this is to create the schema ahead of time and use the `Append` save mode.  This way the schema in Vertica and the schema in your DF can be different, causing rejected rows.  There are examples of this in the functional-tests (see tests where `FaultToleranceTestFail` is true under `EndToEndTests.scala`).

jeremyp-bq:
Also note that the error when the schema in the DF and Vertica differ is the same as ticket #284, where the log indicates that 0 rows were copied but also says that 0 rows were rejected (no data was copied to Vertica, in the target table or the rejects table).

jeremyp-bq:
When testing a DF containing nullable values, against a Vertica table that is not nullable, the error is slightly different.  The logs do report the number of copied rows and rejected rows:
```
21/12/07 22:28:58 WARN SchemaTools: S2V: Column i is NOT NULL in target table "dftest" but it's nullable in the DataFrame. Rows with NULL values in column i will be rejected.
21/12/07 22:28:58 INFO SchemaTools: Load by name. Column list: ("i")
21/12/07 22:28:58 INFO VerticaDistributedFilesystemWritePipe: The copy statement is:
COPY "dftest" ("i") FROM 'webhdfs://hdfs:50070/data/02ac98f0_f957_43a4_8a25_3f975eca117e/*.parquet' ON ANY NODE parquet REJECTED DATA AS TABLE "dftest_02ac98f0_f957_43a4_8a25_3f975eca117e_COMMITS" NO COMMIT
21/12/07 22:28:58 INFO VerticaDistributedFilesystemWritePipe: Performing copy from file store to Vertica
21/12/07 22:28:58 INFO VerticaDistributedFilesystemWritePipe: Checking number of rejected rows via statement: SELECT COUNT(*) as count FROM "dftest_02ac98f0_f957_43a4_8a25_3f975eca117e_COMMITS"
21/12/07 22:28:58 INFO VerticaDistributedFilesystemWritePipe: Verifying rows saved to Vertica is within user tolerance...
[...]

PR #296, opened 2021-12-14
Log rejected data table results

jeremyp-bq:
### Summary

Log rejected data table results

### Description

The rejected data table is never persisted in the Vertica (it is a temporary table), so the user cannot see the exact reason for any rejected rows.  This change logs the first few rows (10) from the rejected data table.  Only the columns that are  most helpful in narrowing down the error are printed (file_name, row_number, rejected_data, and rejected_data_reason).

Trying to persist the table was difficult due to our commit logic, but it can be address as part of #293.

This is a sample of what the rejected data table might contain (not persisted):
```sql
     node_name     |                                     file_name                                     |         session_id          |  transaction_id   | statement_id | batch_number | row_number | rejected_data | rejected_data_orig_length |                    rejected_reason
-------------------+-----------------------------------------------------------------------------------+-----------------------------+-------------------+--------------+--------------+------------+---------------+---------------------------+-------------------------------------------------------
 v_docker_node0001 | webhdfs://hdfs:50070/data/1a4212fb_b2c8_4998_9c1f_09d14b1f9e4a/0-0.snappy.parquet | v_docker_node0001-99:0x57cd | 45035996273707541 |           27 |            0 |          1 | NULL          |                         4 | In column 1: Cannot set NULL value in NOT NULL column
```

And this is a sample of what will now be printed in the logs if there is at least one rejected row (this example has 1 rejected row):
```sh
21/12/14 18:58:13 INFO VerticaDistributedFilesystemWritePipe: Checking number of rejected rows via statement: SELECT COUNT(*) as count FROM "dftest_a3b77254_d039_46cf_98ff_e1cbf81a9c57_COMMITS"
21/12/14 18:58:13 INFO VerticaDistributedFilesystemWritePipe: Verifying rows saved to Vertica is within user tolerance...
21/12/14 18:58:13 INFO VerticaDistributedFilesystemWritePipe: Number of rows_rejected=1. rows_copied=3. failedRowsPercent=0.25. user's failed_rows_percent_tolerance=0.5. passedFaultToleranceTest=true...PASSED.  OK to commit to database.
21/12/14 18:58:13 INFO VerticaDistributedFilesystemWritePipe: Getting rejected rows via statement: SELECT file_name, row_number, rejected_data, rejected_reason FROM "dftest_a3b77254_d039_46cf_98ff_e1cbf81a9c57_COMMITS" LIMIT 10
21/12/14 18:58:13 ERROR VerticaDistributedFilesystemWritePipe: file_name | row_number | rejected_data | rejected_reason
21/12/14 18:58:13 ERROR VerticaDistributedFilesystemWritePipe: webhdfs://hdfs:50070/data/a3b77254_d039_46cf_98ff_e1cbf81a9c57/0-0.snappy.parquet | 1 | NULL | In column 1: Cannot set NULL value in NOT NULL column
```

### Related Issue

#275

### Additional Reviewers

@alexr-bq
@alexey-temnikov
@jonathanl-bq
@ravjotbrar

### DIFF
--- a/connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
+++ b/connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
@@ -367,6 +367,20 @@ class VerticaDistributedFilesystemWritePipe(val config: DistributedFilesystemWri
         logger.info(s"Dropping Vertica rejects table now: " + dropRejectsTableStatement)
         jdbcLayer.execute(dropRejectsTableStatement)
       } else {
+        // Log the first few rejected rows
+        val rejectsDataQuery = "SELECT file_name, row_number, rejected_data, rejected_reason FROM " + rejectsTable + " LIMIT 10"
+        logger.info(s"Getting rejected rows via statement: " + rejectsDataQuery)
+        for {
+          rs <- jdbcLayer.query(rejectsDataQuery)
+          _ = Try {
+            logger.error("file_name | row_number | rejected_data | rejected_reason")
+            while (rs.next) {
+              logger.error(s"${rs.getString(1)} | ${rs.getString(2)} | ${rs.getString(3)} | ${rs.getString(4)}")
+            }
+          }
+          _ = rs.close()
+        } yield ()
+
         Right(())
       }
 
--- a/connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
+++ b/connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
@@ -374,6 +374,7 @@ class VerticaDistributedFilesystemWritePipe(val config: DistributedFilesystemWri
           rs <- jdbcLayer.query(rejectsDataQuery)
           _ = Try {
             val rsmd = rs.getMetaData
+            logger.error(s"Found $rejectedCount rejected rows, displaying up to the first 10:")
             logger.error((1 to rsmd.getColumnCount).map(idx => rsmd.getColumnName(idx)).toList.mkString(" | "))
             while (rs.next) {
               logger.error((1 to rsmd.getColumnCount).map(idx => rs.getString(idx)).toList.mkString(" | "))
--- a/connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
+++ b/connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
@@ -368,13 +368,13 @@ class VerticaDistributedFilesystemWritePipe(val config: DistributedFilesystemWri
         jdbcLayer.execute(dropRejectsTableStatement)
       } else {
         // Log the first few rejected rows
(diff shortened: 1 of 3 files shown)


connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
```diff
@@ -367,6 +367,21 @@ class VerticaDistributedFilesystemWritePipe(val config: DistributedFilesystemWri
         logger.info(s"Dropping Vertica rejects table now: " + dropRejectsTableStatement)
         jdbcLayer.execute(dropRejectsTableStatement)
       } else {
+        // Log the first few rejected rows
+        val rejectsDataQuery = "SELECT file_name, row_number, rejected_data, rejected_reason FROM " + rejectsTable + " LIMIT 10"
```
alexey-temnikov, line 371: If we are using the approach to log 10, shall we try to print different rejected_reason, so we can cover more variations of failures?

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
```diff
@@ -367,6 +367,21 @@ class VerticaDistributedFilesystemWritePipe(val config: DistributedFilesystemWri
...
+        logger.info(s"Getting rejected rows via statement: " + rejectsDataQuery)
```
alexey-temnikov, line 372: I suggest we mention in the logs as well that we are printing only up to 10 rejected rows, so people won't think it is a complete list when reading log files.

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
jeremyp-bq: We could do a GROUP BY on rejected_reason, but then we wouldn't be able to provide the other rows (at least row_number, and possibly file_name and rejected_data as well).

But if we feel the rejected_reason and the count of each distinct reason is more important, we can display that instead.  Having the row_number and rejected_data helps narrow where the rejected row came from, but having a distinct list of rejected_reason and the counts provides a better overview of the rejected rows.

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
jeremyp-bq: Done, see below.

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
jeremyp-bq: Done.

This now provides a good summary of the rejected rows, but also provides example data for each reason (the following example only had a single reason, but the query works and will summarize for up to 10 reasons):
```
21/12/15 22:39:56 ERROR VerticaDistributedFilesystemWritePipe: Found 8 rejected rows, displaying up to 10 of the most common reasons:
21/12/15 22:39:56 ERROR VerticaDistributedFilesystemWritePipe: count | example_data | rejected_reason
21/12/15 22:39:56 ERROR VerticaDistributedFilesystemWritePipe: 8 | NULL | In column 1: Cannot set NULL value in NOT NULL column
```

ravjotbrar (approved):

alexey-temnikov (approved):

Issue #306, opened 2022-01-06
[FEATURE] Support for JDBC Connection Failover


alexey-temnikov:
Currently, there is no mechanism to supply Vertica backup nodes and thus connector can't support scenarios with failover. To address it, a new parameter `backUpServerNodes` needs to be introduced, which will be passed to the JDBC driver (reference: https://www.vertica.com/docs/11.0.x/HTML/Content/Authoring/ConnectingToVertica/ClientJDBC/JDBCConnectionFailover.htm)

alexey-temnikov:
This issue is implemented 3.0.1

PR #307, opened 2022-01-06
[FEATURE] Support for JDBC Connection Failover #306

alexey-temnikov:
### Summary
Currently, there is no mechanism to supply Vertica backup nodes and thus connector can't support scenarios with failover. To address it, a new parameter backupServerNodes needs to be introduced, which will be passed to the JDBC driver (reference: https://www.vertica.com/docs/11.0.x/HTML/Content/Authoring/ConnectingToVertica/ClientJDBC/JDBCConnectionFailover.htm)

### Description
Added new parameter `backupServerNodes`

### Related Issue
https://github.com/vertica/spark-connector/issues/306

### Additional Reviewers
@NerdLogic 
@alexr-bq 
@jonathanl-bq 
@jeremyp-bq


README.md
```diff
@@ -108,44 +108,45 @@ There are a few ways to solve this:
...
+| backUpServerNodes                              | String                                              | The comma separates list of vertica backup nodes. The host name or IP can optionally be followed by a colon and a port number. If not supplied, defaults to the standard Vertica port number (5433). To list multiple hosts, separate them by a comma.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            | No                                            | Empty                       |
```
jeremyp-bq, line 116: I think we should align with all of the other Spark connector properties, which use underscore.  So it should be `backup_server_node` (the JDBC connection property is singular, `BackupServerNode`).  Also note that backup is a single word, so other references in this PR should use `backup` instead of `backUp`.

alexey-temnikov: @jeremyp-bq, good catch! Fixed.

jeremyp-bq (approved):

alexr-bq (approved):

PR #311, opened 2022-01-18
Save job status tables option

### LINKED ISSUE #308: Make creation of job status table optional

Currently the job status table is created on each write to Vertica, if it doesn't already exist.  However this can clutter the database and the option to not create it is desired, or rather don't create it by default and provide an option to create it.

Update the connector to not create the job status table by default.  But also add a new flag, such as `save_metadata_tables`, that will create and write to the status table when enabled.  This flag can also cover the creation of the rejected rows table in the future (see #293).

## Describe alternatives you've considered

We could have the job status table created by default and have the flag to disable that functionality.  However, it seems this table is mostly useful for development/debugging purposes so should probably not create the table by default.

## Additional context

Note that there is an existing `prevent_cleanup` flag to persist temporary HDFS data.  So it makes sense for the new flag to follow this functionality and not persist metadata tables, such as the job status table, by default.

Aryex:
### Summary

Added save_job_tatus_tables:Boolean option for jobs status table.

### Description

With save_job_tatus_tables = true, a job status table will be initialized. Commits to vertica or external tables are recorded to vertica.

### Related Issue

Closes #308 

### Additional Reviewers

@alexey-temnikov
@alexr-bq
@jonathanl-bq
@jeremyp-bq


connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
```diff
@@ -499,6 +503,11 @@ class VerticaDistributedFilesystemWritePipe(val config: DistributedFilesystemWri
 
       _ <- tableUtils.validateExternalTable(config.tablename)
 
+      _ <- if (config.saveMetadataTables) {
+        tableUtils.updateJobStatusTable(config.tablename, config.jdbcConfig.auth.user, 0.0, config.sessionId, success = true)
+      } else {
+        Right(())
+      }
```
jeremyp-bq, line 510: I'm not sure if we should create the status table for external tables, even if the `save_metadata_tables` option is `true`.  Since there is no fault tolerance information saved, the only benefit is to track the number of times an external table is created and the session id it was created under.

@alexey-temnikov, thoughts?

jeremyp-bq:
@Aryex We should also update the README.md to list the new option.

Also note that one of the functional tests is failing (click the details link for the integration tests on the PR, then search the logs for "TEST FAILED" to see the exact one - looks to be an integration test that is looking for the job status table).

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
alexey-temnikov: I agree. We should have it documented and highlighted in the changelog, that this behaviour will be changed. 
@jeremyp-bq, @Aryex, minor nit - should we consider another name instead of `save_metadata_tables` since it is specific to the rejected rows? Perhaps `save_rejected_rows`?

jeremyp-bq: @alexey-temnikov This PR is to turn off the status table by default and add an option to enable it, but I was thinking we could lump together the creation of the rejected rows table in the future under the same option, which is why it has the generic name `save_metadata_tables`.

Unless you think it is better to have an option for creation of each (e.g. `save_status_table` and later `save_rejected_rows_table`)?

Aryex: > I'm not sure if we should create the status table for external tables, even if the `save_metadata_tables` option is `true`. Since there is no fault tolerance information saved, the only benefit is to track the number of times an external table is created and the session id it was created under.
> 
This was purely to mimic the old behavior. That said, it could be extended to provide an option to not record jobs of external tables. Something like: `save_metadata_tables: all | none | vertica-only`.

>This PR is to turn off the status table by default and add an option to enable it, but I was thinking we could lump together the creation of the rejected rows table in the future under the same option. 
>
I think using specific options would be better in the long run?

jeremyp-bq: I was thinking the logic would be like this, unless we see value in creating the status table for external tables:
```
if(config.createExternalTable.isDefined && !config.createExternalTable.isDefined) {
  // create table
} else {
  Right(())
}
```

connector/src/main/scala/com/vertica/spark/config/WriteConfig.scala
```diff
@@ -86,6 +86,7 @@ final case class DistributedFilesystemWriteConfig(jdbcConfig: JDBCConfig,
                                                   failedRowPercentTolerance: Float,
                                                   filePermissions: ValidFilePermissions,
                                                   createExternalTable: Option[CreateExternalTableOption],
+                                                  saveMetadataTables: Boolean = false,
```
jonathanl-bq, line 89: we shouldn't need to do this if the default is false when loading the config anyway

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
Aryex: Sorry, I'm not following the condition. What do you mean by createExternalTable is defined AND not defined?

jeremyp-bq: Typo, fixed it.

To summarize my questions:
1. Should we create the status table for the external table case, or not since it doesn't provide much value for the external tables case (even if the option is set to true)?
2. Should we have separate options for saving the status table and rejected rows table, or use the same option for both?

connector/src/main/scala/com/vertica/spark/config/WriteConfig.scala
Aryex: fixed

jeremyp-bq (approved):

functional-tests/src/main/scala/com/vertica/spark/functests/EndToEndTests.scala
```diff
@@ -1045,6 +1041,49 @@ class EndToEndTests(readOpts: Map[String, String], writeOpts: Map[String, String
...
+    }
+
+    TestUtils.dropTable(conn, tableName)
+  }
+
+  private def getJobStatusTableRowCount(targetTableName:String, stmt:Statement): Int = {
+    val query = "SELECT COUNT (1) FROM S2V_JOB_STATUS_USER_" + writeOpts("user") +
+      " WHERE target_table_name = '" + targetTableName + "'"
+    val rs = stmt.executeQuery(query)
+    var count:Int = 0
```
jonathanl-bq, line 1079: A couple notes here:
1. General style recommendation is to put a space after the colon in type annotations. This is probably something that we should be using an automated code formatter for, but we haven't spent the time to do this yet.
2. Avoid var. In this case, the mutation is localized and probably not a problem, especially since it's in test code, but it helps to get out of the mindset of reaching for var when it isn't necessary. Here, we could use the following snippet instead:
```
val count: Int = if (rs.next) rs.getInt(1) else 0
rs.close()
count
```

jonathanl-bq (approved):

Issue #312, opened 2022-01-27
[BUG] HDFS HA name node is not resolved


jeremyp-bq:
## Environment
- Vertica Spark Connector version: 3.0.1

---

## Problem Description

When using the Spark Connector against HDFS in HA mode if fails to resolve the name node addresses.

For example, consider the following HDFS config:
```
  <property>
    <name>dfs.namenode.http-address.hanameservice.namenode101</name>
    <value>namenode1.domain:9870</value>
  </property>
  <property>
    <name>dfs.namenode.https-address. hanameservice.namenode101</name>
    <value>namenode1.domain:9871</value>
  </property>
```

Results in the following exception:
```
Issues:
Caused by: com.vertica.spark.util.error.ConnectorException: Could not find name node address in Hadoop configuration. Please set either dfs.namenode.http-address or dfs.namenode.https-address in hdfs-site.xml
```

---

jeremyp-bq:
Refer also to the original ticket for HDFS HA support, #248.

jeremyp-bq:
The error indicates that the connector is trying to find the `dfs.namenode.http-address` or `dfs.namenode.https-address` properties in the `hdfs-site.xml` config file.  So it may be a case of not setting the right properties, or not ensuring those properties are available across Vertica, HDFS, and/or the client.

It is important to note that older versions of Hadoop look for `dfs.http.address` and `dfs.https.address` (See deprecated properties list [here](https://hadoop.apache.org/docs/r2.8.2/hadoop-project-dist/hadoop-common/DeprecatedProperties.html)).  So setting those properties as well guards against using older versions of Hadoop.

By commenting out all of the above properties I was able to cause a Kerberos test to fail, although I did not see the exact same error as indicated in this ticket.

Also note that in the example there is a space in the https entry (`dfs.namenode.https-address. hanameservice.namenode101`).  I;m not sure if this would cause any issues or not.

PR #316, opened 2022-02-02
Logging hadoop configs

Aryex:
### Summary

Added debug logs for hadoop configs

### Description

Debug logger will print all hadoop configurations found by Spark when configuring Kerberos.
Hostname and executor id are included.

### Related Issue

#312 

### Additional Reviewers
@alexey-temnikov 
@alexr-bq 
@jeremyp-bq 
@jonathanl-bq


connector/src/main/scala/com/vertica/spark/datasource/jdbc/VerticaJdbcLayer.scala
```diff
@@ -381,7 +383,10 @@ class VerticaJdbcLayer(cfg: JDBCConfig) extends JdbcLayerInterface {
             for {
               nameNodeAddress <- Option(hadoopConf.get(HdfsClientConfigKeys.DFS_NAMENODE_HTTPS_ADDRESS_KEY))
                 .orElse(Option(hadoopConf.get(HdfsClientConfigKeys.DFS_NAMENODE_HTTP_ADDRESS_KEY)))
-                .toRight(MissingNameNodeAddressError())
+                .toRight({
+                  logger.whenDebugEnabled(this.logHadoopConfigs(hadoopConf, session))
+                  MissingNameNodeAddressError()
+                })
```
jeremyp-bq, line 389: I think it would be more useful to always print the Hadoop config above on line 379, if DEBUG is enabled.

examples/kerberos-example/README.md
```diff
@@ -32,6 +25,17 @@ Build the assembly if it hasn't been built already. Note: avoid building it with
...
+If instead you want to use a modified spark connector for this example, create a lib folder at 
+`/kerberos-example` and put your spark connector jar inside. 
+```
+mkdir /spark-connector/examples/kerberos-example/lib
+cp /spark-connector/connector/target/scala-2.12/spark-vertica-connector-assembly-3.0.1.jar /spark-connector/examples/kerberos-example/lib
+```
+Currently, the example uses the latest published spark connector assembly jar as specified in `build.sbt` which 
+it will use instead of the ones in /lib. As such, comment out the line specifying the vertica-spark connector 
+dependency in `build.sbt` so that sbt will look into lib instead. 
+
```
jeremyp-bq, line 38: I think this section needs to be before the Build Example Assembly section.

Aryex: Updated. I also removed the reference to building the connector since we don't need it.

jeremyp-bq (approved):

PR #323, opened 2022-02-09
Use a single JDBC layer per read/write

### LINKED ISSUE #314: [BUG] Spark connector leaves multiple sessions open
## Environment
- Vertica Spark Connector version: 3.0.1

---

## Problem Description

When performing a read or write operation using the Spark Connector, it creates one or more sessions in Vertica.  These sessions are not cleaned up until the connection is closed and the client application exits.

For example, the pyspark example creates 3 sessions, 1 for write and 2 for read.  These remain until the python script completes.

We likely need to close these sessions when an operation completes since we do not appear to reuse a given session id.  That way, even if multiple operations are performed in a single script/client, the sessions should be closed without having to close the client application.

To get the session information, or close all open sessions, run the following:
```
docker exec -it docker_vertica_1 vsql

-- Get all sessions
SELECT session_id, node_name, user_name, client_hostname, login_timestamp, statement_start, current_statement, last_statement, client_type, client_os FROM v_monitor.sessions ORDER BY user_name;

-- Close all sessions
SELECT close_all_sessions();
```

---

jeremyp-bq:
### Summary

Use a single JDBC layer, which holds the actual connection, for read or write operations.

### Description

We create a new JDBC layer, and thus a new connection, each time a read or write pipe is called for.  This results in additional connections at runtime, some of which are not closed when the operation completes.  This fix recreates the JDBC layer each time it is called, if the underlying connection was closed.

Further work is still needed to avoid opening and closing multiple connections, which will be performed under another ticket.

### Related Issue

Closes #314.

### Additional Reviewers

@alexey-temnikov
@alexr-bq
@Aryex
@jonathanl-bq

### DIFF
--- a/connector/src/main/scala/com/vertica/spark/datasource/core/factory/VerticaPipeFactory.scala
+++ b/connector/src/main/scala/com/vertica/spark/datasource/core/factory/VerticaPipeFactory.scala
@@ -35,7 +35,12 @@ trait VerticaPipeFactoryInterface {
 /**
  * Implementation of the vertica pipe factory
  */
-object VerticaPipeFactory extends VerticaPipeFactoryInterface{
+object VerticaPipeFactory extends VerticaPipeFactoryInterface {
+
+  // Maintain a single copy of the read and write JDBC layers
+  private var readLayer: Option[VerticaJdbcLayer] = None;
+  private var writeLayer: Option[VerticaJdbcLayer] = None;
+
   override def getReadPipe(config: ReadConfig): VerticaPipeInterface with VerticaPipeReadInterface = {
     config match {
       case cfg: DistributedFilesystemReadConfig =>
@@ -47,18 +52,34 @@ object VerticaPipeFactory extends VerticaPipeFactoryInterface{
           }
           case _ => None
         })
+        val jdbcLayer = readLayer match {
+          case Some(readLayer) => readLayer
+          case None => {
+            readLayer = Some(new VerticaJdbcLayer(cfg.jdbcConfig))
+            readLayer.get
+          }
+        }
+        // TODO: Reopen the connection if it is closed
         new VerticaDistributedFilesystemReadPipe(cfg, hadoopFileStoreLayer,
-          new VerticaJdbcLayer(cfg.jdbcConfig),
+          jdbcLayer,
           new SchemaTools,
           new CleanupUtils
         )
     }
   }
+
   override def getWritePipe(config: WriteConfig): VerticaPipeInterface with VerticaPipeWriteInterface = {
     config match {
       case cfg: DistributedFilesystemWriteConfig =>
         val schemaTools = new SchemaTools
-        val jdbcLayer = new VerticaJdbcLayer(cfg.jdbcConfig)
+        val jdbcLayer = writeLayer match {
+          case Some(writeLayer) => writeLayer
+          case None => {
+            writeLayer = Some(new VerticaJdbcLayer(cfg.jdbcConfig))
+            writeLayer.get
+          }
+        }
+        // TODO: Reopen the connection if it is closed
         new VerticaDistributedFilesystemWritePipe(cfg,
           new HadoopFileStoreLayer(cfg.fileStoreConfig, Some(cfg.schema)),
           jdbcLayer,
@@ -67,4 +88,5 @@ object VerticaPipeFactory extends VerticaPipeFactoryInterface{
         )
     }
   }
+
 }
--- a/connector/src/main/scala/com/vertica/spark/datasource/core/factory/VerticaPipeFactory.scala
+++ b/connector/src/main/scala/com/vertica/spark/datasource/core/factory/VerticaPipeFactory.scala
@@ -53,13 +53,17 @@ object VerticaPipeFactory extends VerticaPipeFactoryInterface {
           case _ => None
         })
         val jdbcLayer = readLayer match {
-          case Some(readLayer) => readLayer
+          case Some(layer) => {
+            if (layer.isClosed()) {
+              readLayer = Some(new VerticaJdbcLayer(cfg.jdbcConfig))
+            }
+            readLayer.get
+          }
           case None => {
             readLayer = Some(new VerticaJdbcLayer(cfg.jdbcConfig))
             readLayer.get
           }
         }
-        // TODO: Reopen the connection if it is closed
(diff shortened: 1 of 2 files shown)


connector/src/main/scala/com/vertica/spark/datasource/core/factory/VerticaPipeFactory.scala
```diff
@@ -35,7 +35,12 @@ trait VerticaPipeFactoryInterface {
 /**
  * Implementation of the vertica pipe factory
  */
-object VerticaPipeFactory extends VerticaPipeFactoryInterface{
+object VerticaPipeFactory extends VerticaPipeFactoryInterface {
+
+  // Maintain a single copy of the read and write JDBC layers
+  private var readLayer: Option[VerticaJdbcLayer] = None;
+  private var writeLayer: Option[VerticaJdbcLayer] = None;
```
jeremyp-bq, line 42: I was unable to create a singleton of the connection or JDBC layer due to our use of a lazy connection and the fact the JDBC layer may contain a closed connection (cannot be reopened, must create a new one).

These vars hold a single JDBC layer per operation (read and write), and the logic below will recreate them (and their underlying connection) if the connection was closed.

In the future we should properly refactor our use of the JDBC layer and connection to avoid closing and opening connections unnecessarily (will create a separate ticket for that).

connector/src/main/scala/com/vertica/spark/datasource/core/factory/VerticaPipeFactory.scala
```diff
@@ -47,18 +52,42 @@ object VerticaPipeFactory extends VerticaPipeFactoryInterface{
...
+            if (layer.isClosed()) {
+              writeLayer = Some(new VerticaJdbcLayer(cfg.jdbcConfig))
+            }
+            writeLayer.get
+          }
+          case None => {
+            writeLayer = Some(new VerticaJdbcLayer(cfg.jdbcConfig))
+            writeLayer.get
+          }
+        }
```
Aryex, line 90: Duplicated logic like above

connector/src/main/scala/com/vertica/spark/datasource/core/factory/VerticaPipeFactory.scala
```diff
@@ -47,18 +52,42 @@ object VerticaPipeFactory extends VerticaPipeFactoryInterface{
...
+            if (layer.isClosed()) {
+              readLayer = Some(new VerticaJdbcLayer(cfg.jdbcConfig))
+            }
+            readLayer.get
+          }
+          case None => {
+            readLayer = Some(new VerticaJdbcLayer(cfg.jdbcConfig))
+            readLayer.get
+          }
+        }
```
Aryex, line 66: I think  we can refactor this duplicated logic:
```
  private def getJdbcLayer(jdbcConfig: JDBCConfig) = {
    val jdbcLayer = readLayer match {
      case Some(layer) => {
        if (layer.isClosed()) {
          readLayer = Some(new VerticaJdbcLayer(jdbcConfig))
        }
        readLayer.get
      }
      case None => {
        readLayer = Some(new VerticaJdbcLayer(jdbcConfig))
        readLayer.get
      }
    }
    jdbcLayer
  }
```

Aryex (approved):

connector/src/main/scala/com/vertica/spark/datasource/core/factory/VerticaPipeFactory.scala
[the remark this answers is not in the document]
jeremyp-bq: Hard to break that muscle memory, will fix.

connector/src/main/scala/com/vertica/spark/datasource/core/factory/VerticaPipeFactory.scala
```diff
@@ -35,7 +35,12 @@ trait VerticaPipeFactoryInterface {
...
-object VerticaPipeFactory extends VerticaPipeFactoryInterface{
+object VerticaPipeFactory extends VerticaPipeFactoryInterface {
+
+  // Maintain a single copy of the read and write JDBC layers
+  private var readLayer: Option[VerticaJdbcLayer] = None;
+  private var writeLayer: Option[VerticaJdbcLayer] = None;
+
   override def getReadPipe(config: ReadConfig): VerticaPipeInterface with VerticaPipeReadInterface = {
     config match {
       case cfg: DistributedFilesystemReadConfig =>
```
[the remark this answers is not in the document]
jeremyp-bq: Looks like there is only one class that extends ReadConfig at the moment, and that is DistributedFilesystemReadConfig.  Same applies to the write case.  So I don't think it is needed at this time (and I don't want to expand the scope of this ticket).

connector/src/main/scala/com/vertica/spark/datasource/core/factory/VerticaPipeFactory.scala
Aryex: Sorry, I see the mistake in my code. 
Would something like this work?
```
  readLayer = checkJdbcLayer(readLayer,cfg.jdbcConfig)
  new VerticaDistributedFilesystemReadPipe(cfg, hadoopFileStoreLayer,
    readLayer.get,
    new SchemaTools,
    new CleanupUtils
  )
```
```
writeLayer = checkJdbcLayer(writeLayer, cfg.jdbcConfig)
new VerticaDistributedFilesystemWritePipe(cfg,
  new HadoopFileStoreLayer(cfg.fileStoreConfig, Some(cfg.schema)),
  writeLayer.get,
  schemaTools,
  new TableUtils(schemaTools, writeLayer.get)
)
```
``` 
  private def checkJdbcLayer(jdbcLayer: Option[VerticaJdbcLayer], jdbcConfig: JDBCConfig): Option[VerticaJdbcLayer] = jdbcLayer match {
      case Some(layer) => if (layer.isClosed()) Some(new VerticaJdbcLayer(jdbcConfig)) else jdbcLayer
      case None => Some(new VerticaJdbcLayer(jdbcConfig))
  }
```

jeremyp-bq: Sure, we can go with that for now.  Eventually this will need to be refactored again as we are still opening and closing multiple connections per read or write, but that is a larger effort.

alexr-bq (approved):

PR #325, opened 2022-02-10
Add an integration test for session management

jeremyp-bq:
### Summary

Add an integration test for session management.

### Description

Check that sessions are cleaned up after a read or write operation.

### Related Issue

#323

### Additional Reviewers

@alexey-temnikov 
@alexr-bq 
@Aryex 
@jonathanl-bq


connector/src/main/scala/com/vertica/spark/datasource/core/DSWriter.scala
```diff
@@ -94,6 +94,9 @@ class DSWriter(config: WriteConfig, uniqueId: String, pipeFactory: VerticaPipeFa
   }
 
   def commitRows(): ConnectorResult[Unit] = {
-    pipe.commit()
+    val ret = pipe.commit()
+    // Ensure all connections are closed, including read connections used by the write operation
+    val _ = pipeFactory.closeJdbcLayers()
```
jeremyp-bq, line 99: After testing a few more permutations I found that a write (without a read) created a read connection for getting schema/column information from Vertica.  The read connection was not being closed, resulting in a dangling session in that case.

functional-tests/src/main/scala/com/vertica/spark/functests/EndToEndTests.scala
```diff
@@ -4129,5 +4131,39 @@ class EndToEndTests(readOpts: Map[String, String], writeOpts: Map[String, String
...
+    val mode = SaveMode.Overwrite
+
+    for (i <- 1 to 10) {
+      println("Performing multiple writes and reads - iteration " + i)
+      df.write.format("com.vertica.spark.datasource.VerticaSource").options(writeOpts + ("table" -> tableName)).mode(mode).save()
+      val dfRead: DataFrame = spark.read.format("com.vertica.spark.datasource.VerticaSource").options(readOpts + ("table" -> tableName)).load()
+      assert(dfRead.count() == 3)
+    }
+    // Ensure operations have time to complete and cleanup
+    Thread.sleep(5000)
```
jeremyp-bq, line 4149: Didn't want to do this, but sometimes it failed on GitHub, and other times it didn't.  I never had to sleep when running the tests locally.

functional-tests/src/main/scala/com/vertica/spark/functests/EndToEndTests.scala
```diff
@@ -4129,5 +4131,45 @@ class EndToEndTests(readOpts: Map[String, String], writeOpts: Map[String, String
     fsLayer.createDir(fsConfig.address, "777")
   }
 
+  // Ignore test for now as it sometimes fails on GitHub
```
Aryex, line 4134: Should we mention this in the functional test's readme?

Aryex (approved):

functional-tests/src/main/scala/com/vertica/spark/functests/EndToEndTests.scala
jeremyp-bq: I don't think it is necessary, it would require updating the code then running the tests, which is messy.  Ideally we would use scalatest tags via sbt testOnly so a dev could include this test is a given run, but our current integration test framework does not work with sbt testOnly.

Session lifecycles in the integration tests, especially in Docker when run by GitHub actions, are not consistent enough to test.  Although I'm hoping we can revisit this test in the future if we find a more consistent way of verifying sessions in GitHub.

alexr-bq (approved):

Issue #326, opened 2022-02-11
[FEATURE] Support pushdown aggregate for Spark 3.2


Aryex:
## Description
Implement aggregate pushdown interface which was introduced in Spark 3.2.
The connector should pushdown `groupBy` as well as aggregates defined supported by Spark pushdown, which includes:
- Count
- Min
- Max
- Sum

PR #327, opened 2022-02-11
Pushdown aggregate

### LINKED ISSUE #319: [Investigation] How the connector interacts with a dataframe.count()
Currently, a count operation would require invoke a read which is costly compared to a JDBC command. If the connector can detect that a count is part of the execution plan, it could use jdbc instead. 

This could be applied to other simple spark functions as well.

Aryex:
### Summary

Add push down support for count

### Description
Implemented pushdown aggregate support for count operations.
Aggregate pushdown was first introduced by Spark in 3.2 and thus is not available if you use an older spark version. 

### Related Issue
Closes #319 
#326

### Additional Reviewers
@alexey-temnikov 
@alexr-bq 
@jeremyp-bq 
@jonathanl-bq

### DIFF
--- a/connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemReadPipe.scala
+++ b/connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemReadPipe.scala
@@ -185,13 +185,13 @@ class VerticaDistributedFilesystemReadPipe(
 
   private def getSelectClause: ConnectorResult[String] = {
       val pushdownAggregations = this.config.getPushdownAggregation.aggregateExpressions()
-      if(pushdownAggregations.isEmpty) getColumnNames(this.config.getRequiredSchema) else this.getAggregateColumns
+      if(pushdownAggregations.isEmpty) getColumnNames(this.config.getRequiredSchema) else this.getPushdownAggregateColumns
   }
 
-  private def getAggregateColumns: ConnectorResult[String] = {
+  private def getPushdownAggregateColumns: ConnectorResult[String] = {
     val pushdownAggregates = this.config.getPushdownAggregation.aggregateExpressions
     val aggregateSelectClause = pushdownAggregates.map {
-      case agg: CountStar => agg.toString + " as \"count(a)\""
+      case agg: CountStar => agg.toString + " as \"" + this.config.getRequiredSchema.fields.head + "\""
       case agg: Count => agg.toString +  " as \"" + agg.column.describe + "\""
       case _ => ""
     }.mkString(", ")
--- a/connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemReadPipe.scala
+++ b/connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemReadPipe.scala
@@ -241,7 +241,7 @@ class VerticaDistributedFilesystemReadPipe(
       // File permissions.
       filePermissions = config.filePermissions
 
-      cols <- getColumnNames(this.config.getRequiredSchema)
+      cols <- if(config.getPushdownCount()) Right(s"COUNT(*)") else getColumnNames(this.config.getRequiredSchema)
       _ = logger.info("Columns requested: " + cols)
 
       pushdownSql = this.addPushdownFilters(this.config.getPushdownFilters)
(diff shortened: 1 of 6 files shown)


examples/basic-read-example/src/main/scala/example/Main.scala
```diff
@@ -27,7 +27,8 @@ object Main {
       "user" -> conf.getString("functional-tests.user"),
       "db" -> conf.getString("functional-tests.db"),
       "staging_fs_url" -> conf.getString("functional-tests.filepath"),
-      "password" -> conf.getString("functional-tests.password")
+      "password" -> conf.getString("functional-tests.password"),
```
jeremyp-bq, line 30: This probably belongs under the demo example.  There are already cases for column and filter pushdown in there.

Aryex: This file wasn't supposed to be committed. I'll revert it.

connector/src/main/scala/com/vertica/spark/datasource/v2/VerticaDatasourceV2Read.scala
```diff
@@ -19,10 +19,16 @@ import org.apache.spark.sql.types._
...
 import com.vertica.spark.datasource.core.{DSConfigSetupInterface, DSReader, DSReaderInterface}
+import com.vertica.spark.util.error.ErrorHandling.ConnectorResult
 import com.vertica.spark.util.error.{ConnectorError, ErrorHandling, InitialSetupPartitioningError}
 import com.vertica.spark.util.pushdown.PushdownUtils
+import org.apache.spark.sql.catalyst.expressions.aggregate.AggregateFunction
+import org.apache.spark.sql.connector.expressions.NamedReference
+import org.apache.spark.sql.connector.expressions.aggregate.{AggregateFunc, Aggregation, Count, CountStar}
 import org.apache.spark.sql.sources.Filter
 
+import scala.::
```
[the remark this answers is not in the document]
Aryex: fixed

connector/src/main/scala/com/vertica/spark/datasource/v2/VerticaDatasourceV2Read.scala
```diff
@@ -31,22 +32,40 @@ case class PushFilter(filter: Filter, filterString: String) extends PushdownFilt
...
+case class UnknownColumnName(colName: String) extends ConnectorError {
+  def getFullContext: String = s"$colName could not be found"
+}
+
 /**
   * Builds the scan class for use in reading of Vertica
   */
 class VerticaScanBuilder(config: ReadConfig, readConfigSetup: DSConfigSetupInterface[ReadConfig]) extends ScanBuilder with
-  SupportsPushDownFilters with SupportsPushDownRequiredColumns {
+  SupportsPushDownFilters with SupportsPushDownAggregates with SupportsPushDownRequiredColumns {
```
jeremyp-bq, line 57: Note that our functional tests are failing for versions other than 3.2.0 (Spark 3.0.2 and 3.1.1) since SupportsPushDownAggregates is only supported in 3.2.0.  Unfortunately our functional tests are still marked as a success even though the EndToEndTests fail to run.

So we need to determine how to best support old and new versions before this is merged (perhaps add a Trait at runtime if aggregates are supported?).

I will make a ticket in the backlog to address how running functional tests that fail are still marked as success.

Aryex: resolved with alex's merge

jeremyp-bq (approved):
Tests for all versions of Spark are now passing, looks good to me.

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemReadPipe.scala
```diff
@@ -259,8 +270,7 @@ class VerticaDistributedFilesystemReadPipe(
         ", rowGroupSizeMB = " + maxRowGroupSize +
         ", fileMode = '" + filePermissions +
         "', dirMode = '" + filePermissions +
-        "') AS " + "SELECT " + cols +
-        " FROM " + exportSource + pushdownSql + ";"
+        "') AS SELECT " + selectClause + " FROM " + exportSource + pushdownFilters + groupbyClause + ";"
```
alexr-bq, line 273: I was envisioning that we'd skip the export altogether, but thinking about it that would require some significant extra effort. Have you tested the performance of a count() using this method vs previous? What are the results of that?

Aryex: Just did a test on the docker environment counting a table of 100,000 rows. 

- With pushdown: vertica export took 127ms, read took 818ms

- Without pushdown vertica export took 600ms, and reading parquets took ~1400ms x 6 parquets. 

Note that this is on the docker environment so I think it would be worth it to test this on an actual cluster. Still I would expect this implementation to be faster on larger set of data since we would be making one small export instead of multiple exports. 

As for the idea of skipping the export, I think it is doable if we: 

1. query instead of export
2. return partition count to 1 and pass query result to worker 
3. worker construct the data instead of reading partition. 

1 would be replacing the export sql query to a normal query then return only a single partition input, 2 and 3 would require us to look into how the InputPartition data are serialized to workers to see if we could encode some data to it, OR we can construct a special factory that construct a reader who has the data with it already.

alexr-bq (approved):

PR #332, opened 2022-02-18
aws_enable_ssl behavior prior to Vertica 11.

### LINKED ISSUE #321: [FEATURE] Prior to vertica 11 doesn't have session level parameter for AWSEnableHttps.
Spark connector V2 try to alter session for AWSEnableHttps but prior to vertica 11 doesn't support session level parameter for AWSEnableHttps.
So this connector doesn't work for prior to vertica 11 and object storage that is not configured ssl mode.

vertica=> select version(), parameter_name, allowed_levels from configuration_parameters where parameter_name = 'AWSEnableHttps';
               version               | parameter_name | allowed_levels 
-------------------------------------+----------------+----------------
 Vertica Analytic Database v10.1.1-6 | AWSEnableHttps | DATABASE

Aryex:
### Summary

Updated readme to reflect aws_enable_ssl behavior prior to Vertica 11.

### Description

It seems that Vertica versions before 11, we cannot set SSL at a session level and must be enabled on Vertica. In Vertica 11, the connector can do this only if the user is admin. 

### Related Issue

Closes #321

### Additional Reviewers
@alexey-temnikov 
@alexr-bq 
@jeremyp-bq 
@jonathanl-bq


jeremyp-bq:
> What is the error message that the user will get when trying these bad cases? If it's unclear can we wrap it in something more clear?

It turns out that the AWSEnableHttps parameter is only supported at a database level in 10.x and before, but 11.x supports it at a session level as well as at a database level.  At the moment the connector simply fails on older versions this this is set since it tries to set it at a session level.

So there are actually some code changes required to only set it at a session level if Vertica supports it, otherwise do not set it (and the user will have to set it at a database level outside the connector, as described in the README).  Alex L is working on that now.

To your point we could probably add a warning in the log if someone sets this option on older versions of Vertica (since it will not do anything).

connector/src/main/scala/com/vertica/spark/datasource/jdbc/VerticaJdbcLayer.scala
```diff
@@ -321,9 +321,26 @@ class VerticaJdbcLayer(cfg: JDBCConfig) extends JdbcLayerInterface {
     } yield ()
   }
 
+  def getVersionNumber(): ConnectorResult[Int] = {
+    this.query("SELECT version();") match {
+      case Right(rs) => {
+        rs.next()
+        val versionStr = rs.getString(1)
+        // Regex pattern to extract version number.
+        val pattern = ".*v([0-9]+)\\.[0-9]+\\.[0-9]+-[0-9]+.*".r
```
Aryex, line 330: I'm not sure if regex should be used here, or simple string manipulations is enough?

jeremyp-bq: I think regex is fine, looks like all versions of Vertica use semantic versioning.

Although we might want to consider wrapping this whole block in a try and returning a default version (11?) if the select query or the regex fail to get the version.  That would be similar to what Alex did when getting the Spark version.

We should also consider updating this method to return the entire version string instead of just the major number (the caller would then have to split and compare the part(s) they are interested in).

Aryex: Updated in the latest push. It now defaults to the latest Vertica major version 11.0

connector/src/main/scala/com/vertica/spark/datasource/jdbc/VerticaJdbcLayer.scala
```diff
@@ -451,3 +481,15 @@ class VerticaJdbcLayer(cfg: JDBCConfig) extends JdbcLayerInterface {
...
+
+object VerticaVersion{
+  def make(str: String): VerticaVersion = {
+    val pattern = ".*v([0-9]+\\.[0-9]+)\\.([0-9])+-([0-9]+).*".r
+    val pattern(major, service, hotfix) = str
+    new VerticaVersion(major.toDouble, service.toInt, hotfix.toInt)
+  }
+}
+case class VerticaVersion(majorReleaseNumber: Double, serviceReleaseNumber: Int, hotfixNumber: Int){
+  override def toString: String = s"${majorReleaseNumber}.${serviceReleaseNumber}-$hotfixNumber"
```
jeremyp-bq, line 493: I believe it is `${major}.${minor}.${service-pack}-${hotfix}`.  e.g. `11.1.0-0`, where `11` is the major and `1` is the minor (no service pack or hotfix yet, so both `0`).

Aryex: I'll add it to the next push. I guess I was following this vertica article on their version number. 
https://www.vertica.com/blog/version-vertica-running/

jeremyp-bq: That post is not very clear, it seems to merge the major and minor together.  I have seen better documentation on the Vertica Confluence (internal).

There is some info on the Vertica doc site, but no specific page on versioning: 
* https://www.vertica.com/docs/11.0.x/HTML/Content/Authoring/InstallationGuide/Upgrade/UpgradePaths.htm
* https://www.vertica.com/docs/11.0.x/HTML/JavaSDK/classcom_1_1vertica_1_1sdk_1_1_build_info.htm

PR #351, opened 2022-03-24
Vertica set write

Aryex:
### Summary

Implantation for supporting writing SET to Vertica.

### Description

Arrays can be marked in it's metadata as a SET. Because of how SET works in Vertica, we should have the following behavior.

- When writing a new table to Vertica, a SET column will be created.
- When overwriting a table in Vertica, the table will be recreated with a SET column.
- When appending to Vertica, we can append to both SET and ARRAY.

### Additional Reviewers

@alexey-temnikov 
@alexr-bq 
@jeremyp-bq 
@jonathanl-bq


examples/complex-data-types/array-example/src/main/scala/example/Main.scala
```diff
@@ -1,117 +0,0 @@
...
-  }
-
-  def save1DArrayToVertica(spark: SparkSession, options: Map[String, String]): Unit = {
-    val tableName = "1D_array_test"
-    // Table name needs to be specified in option
-    val writeOpts = options + ("table" -> tableName)
-    // Define schema of a table with a 1D array column
-    val schema = new StructType(Array(StructField("col1", ArrayType(IntegerType))))
-    // data
-    val data = Seq(Row(Array(1, 2, 3, 4, 5, 6)))
```
jeremyp-bq, line 82: Shouldn't we keep the array example?

And maybe the array example should have duplicate elements (to contrast with set, which will remove duplicates).

jonathanl-bq (approved):

examples/complex-data-types/array-example/src/main/scala/example/Main.scala
Aryex: I have no idea why this was removed lol. Re-added and updated.

.gitignore
```diff
@@ -108,3 +108,4 @@ jitwatch.out
 .metals
 .bloop
 project/**/metals.sbt
+#
```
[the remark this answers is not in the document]
Aryex: Fixed

PR #353, opened 2022-03-24
Release 3.1.0

Aryex:
### Summary

PR for release 3.1.0

### Description

- Updated version numbers.
- Examples' build.sbt and shell scripts now pull version numbers from a properties file

@alexey-temnikov 
@alexr-bq 
@jeremyp-bq 
@jonathanl-bq


CONTRIBUTING.md
```diff
@@ -231,7 +231,7 @@ The functional-tests project is meant to run against a real environment. See "Se
 
 This project contains a series of end-to-end tests. It also contains tests for components on the bottom layer of the connector, those that do some kind of IO directly, and thus don't make as much sense to unit test.
 
-If you set the sparkVersion in build.sbt to 3.0.3, you will also need to use hadoop-hdfs version 2.7.0 when running `sbt run` to run the integration tests.
+If you set the sparkVersion in build.sbt to 3.0.4, you will also need to use hadoop-hdfs version 2.7.0 when running `sbt run` to run the integration tests.
```
jeremyp-bq, line 234: This looks to be the Spark version, not the connector version.  It looks to have been caught up in find and replace by accident in the past as well,  think it should be 3.0.0 here.

examples/examples.properties
```diff
@@ -0,0 +1 @@
+connector-version:3.0.4
```
jeremyp-bq, line 1: If we add this file we should add it to the connector itself so it is a single source of truth that be read by the connector itself, the examples, and the functional tests.

There is a https://github.com/sbt/sbt-release project that looks for the version in a version.sbt file, although not sure if it is too complicated (as we just want a common location for the project version, we don't need to do any further release work).

CONTRIBUTING.md
Aryex: fixed.

examples/examples.properties
Aryex: Changed. The version file is now at project root

jeremyp-bq (approved):

PR #336, opened 2022-02-25
Array support

Aryex:
### Summary
Implementation for initial support of Arrays
### Description
- On write operations, nested array is supported
- On read operations, only 1D array is supported
- While array can support complex type elements, only array is implemented at the moment.
- Added 'arrlen' option to set array length on write. arrlen specify the number of elements in all arrays. If equals 0 then no length is specified to Vertica. The default arrlen 0. 
- Updated unit tests, functional tests, and readme accordingly
### Notable changes:
- Main changes are in SchemaTools with code added to handle array type.
- getColumnInfo() was mainly used on read to get column types. This was updated to support array type. A good starting point for looking at read logic for handling arrays. 
- getVerticaTypeFromSparkType() is mainly used on writes to Vertica. Good starting point for looking at write logic.

### Related Issue

#337 
Closes #337 
Closes #339 

### Additional Reviewers
@alexey-temnikov 
@alexr-bq 
@jeremyp-bq 
@jonathanl-bq

PR #357, opened 2022-03-29
Row write support

### LINKED ISSUE #343: [FEATURE] Support Row write
## Description
Implement writing Spark sql Structs as Vertica rows

Aryex:
### Summary
Added row write support 

### Description

SchemaTools can now parse Spark struct to Vertica Rows
Added support for Arrays[Rows]
Examples to be added in another PR

### Related Issue

Closes #343

### Additional Reviewers

@alexey-temnikov 
@alexr-bq 
@jeremyp-bq 
@jonathanl-bq


alexr-bq (approved):

connector/src/main/scala/com/vertica/spark/util/error/ErrorHandling.scala
```diff
@@ -472,6 +472,11 @@ case class ParamsNotSupported(operation: String) extends JdbcError {
   */
 trait SchemaError extends ConnectorError
 
+case class StructFieldsError(error: ConnectorError) extends SchemaError {
+  def getFullContext: String = s"${this.error.getClass.getSimpleName}: ${error.getFullContext}"
+  override def getUserMessage: String = s"${this.error.getClass.getSimpleName}: ${error.getFullContext}"
```
jonathanl-bq, line 477: Did this need to be overridden? I think the default is to just call getFullContext anyway, which this appears to basically be doing.

connector/src/main/scala/com/vertica/spark/util/schema/SchemaTools.scala
```diff
@@ -399,13 +399,25 @@ class SchemaTools extends SchemaToolsInterface {
...
-           org.apache.spark.sql.types.StructType(_) => Right("VARBINARY(" + longlength + ")")
+      case org.apache.spark.sql.types.MapType(_,_,_) => Right("VARBINARY(" + longlength + ")")
+      case org.apache.spark.sql.types.StructType(fields) => sparkStructToVerticaRow(fields, strlen, arrayLength)
       case org.apache.spark.sql.types.ArrayType(sparkType,_) => sparkArrayToVerticaArray(sparkType, strlen, arrayLength)
       case _ => this.sparkPrimitiveToVerticaPrimitive(sparkType, strlen)
     }
   }
 
+  private def sparkStructToVerticaRow(fields: Array[StructField], strlen: Long, arrayLength: Long): SchemaResult[String] = {
+    makeTableColumnDefs(StructType(fields), strlen, null, arrayLength) match {
```
jonathanl-bq, line 410: Why are we passing in null here?

jonathanl-bq (changes requested):

connector/src/main/scala/com/vertica/spark/util/error/ErrorHandling.scala
Aryex: Good call. I was copying from other errors so I didn't realized it default to getFullContext.

connector/src/main/scala/com/vertica/spark/util/schema/SchemaTools.scala
Aryex: Fixed. The argument was for a jdbc layer but it is not being used at all. It was added from PR #336 so must've been an oversight on my end.

jonathanl-bq (approved):

PR #359, opened 2022-03-30
Turn off codecov patch check

Aryex:
### Summary

Updated codecov.yml to turn off patch status check on each PR. Patch status only looks at the coverage of the PR it self.
https://docs.codecov.com/docs/commit-status#patch-status

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -5,3 +5,4 @@ coverage:
         # basic
         target: 80
         threshold: 1%
+    patch: off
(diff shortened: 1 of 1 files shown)


alexr-bq:
Why are we removing this?

Aryex:
> Why are we removing this?

I'm thinking this is unneeded? It checks how much the changes made by a PR are covered. Right now it's set to 100% but I could reduce it to 80% as well.

jeremyp-bq (approved):

Issue #360, opened 2022-03-31
Complex types functional testing for Vertica 10


Aryex:
Vertica 10 only supports a subset of complex types features. To support V10 users, we should include regression testing for V10.

Aryex:
This includes creating a nightly test to execute the test that we want.

PR #369, opened 2022-04-14
Parameterize functional tests

Aryex:
### Summary

Functional tests driver application now takes optional arguments

### Description

- Introduced abstract class EndToEnd which initializes variables for testing operations between Spark and Vertica
- Complex type functional tests are split into a separate class.
- Uses [scopt](https://github.com/scopt/scopt) to parse command-line arguments.
  - With no arguments, the default set of tests are executed
  - Use `sbt run -h` for a list of options. 

### Related Issue

#360 

### Additional Reviewers

@alexey-temnikov 
@alexr-bq 
@jonathanl-bq 
@jeremyp-bq


functional-tests/README.md
```diff
@@ -25,7 +25,11 @@ In the sandbox environment, change your working directory to functional-tests
 cd spark-connector/functional-tests
 ```
 
-From the functional-tests directory, `sbt run` from the command line. Alternatively, to run the LargeDataTests, use `sbt "run Large"`.
+### Starting Functional Tests
+
+To execute the default functional tests, use `sbt run` from the command line.
+
+By default, a set of functional test suites will be executed, but you can specify arguments to modify you test run. For example, `sbt run -l` will add LargeDataTests to the test suites for execution. Use `sbt run -h` for a list of arguments.
```
alexey-temnikov, line 32: Can we add few examples?

Aryex: Updated

functional-tests/src/main/scala/Main.scala
```diff
@@ -121,129 +124,213 @@ object Main extends App {
...
+
+  private def defaultTestSuites: String = {
+    val result = Seq(
+      new JDBCTests(jdbcConfig),
+      new HDFSTests(fileStoreConfig, jdbcConfig),
+      new CleanupUtilTests(fileStoreConfig),
+      new EndToEndTests(readOpts, writeOpts, jdbcConfig, fileStoreConfig),
+      new ComplexTypeTests(readOpts, writeOpts, jdbcConfig, fileStoreConfig)
+    ).mkString("\n")
+    closeSparkSession()
```
jeremyp-bq, line 236: Do we need to close the session here?  If so, maybe add a comment as to why.

Aryex: Fixed. We should now not need to close spark session there anymore. It was caused by HDFSTests creating a session on init but never closing it.

jeremyp-bq (approved):

PR #373, opened 2022-04-19
Adding v10 functional tests and fix SchemaToolsV10 not detecting string types

Aryex:
### Summary

Adding functional tests specific to Vertica 10 complex types operations.
Fixed an issue with SchemaToolsV10 not properly detecting string type.

### Related Issue

#360 

### Additional Reviewers

@alexey-temnikov 
@alexr-bq 
@jeremyp-bq 
@jonathanl-bq


connector/src/main/scala/com/vertica/spark/util/schema/SchemaTools.scala
```diff
@@ -21,7 +21,7 @@ import com.vertica.spark.util.complex.ComplexTypeUtils
 import com.vertica.spark.util.error.ErrorHandling.{ConnectorResult, SchemaResult}
 import com.vertica.spark.util.error._
 import com.vertica.spark.util.schema.SchemaTools.{VERTICA_NATIVE_ARRAY_BASE_ID, VERTICA_PRIMITIVES_MAX_ID, VERTICA_SET_BASE_ID, VERTICA_SET_MAX_ID}
-import org.apache.spark.sql.types._
+import org.apache.spark.sql.types.{StringType, _}
```
jeremyp-bq, line 24: Should either use just the wildcard or expand all the types (I assume there are a lot, which is why just the wildcard was specified before).

connector/src/main/scala/com/vertica/spark/util/schema/SchemaTools.scala
```diff
@@ -761,14 +761,21 @@ class SchemaToolsV10() extends SchemaTools {
...
-        .left.map(errors => ErrorList(errors))
+        colList.map(col => checkColumnIsCT(col, tableInfo.tableName, tableInfo.dbSchema, jdbcLayer))
+          .toList
+          .traverse(_.leftMap(err => NonEmptyList.one(err)).toValidated).toEither
+          .map(list => list)
+          .left.map(errors => ErrorList(errors))
+    }
+  }
+
+  private def checkColumnIsCT(col: ColumnDef, tableName: String, dbSchema: String, jdbcLayer: JdbcLayerInterface): ConnectorResult[ColumnDef] = {
```
jeremyp-bq, line 772: CT is not immediately obvious, we deal with complex types and catalyst types in this class.  We should get in the habit of adding scaladocs to all methods (new methods and any existing methods we work on).

connector/src/main/scala/com/vertica/spark/util/schema/SchemaTools.scala
Aryex: Fixed. I'm guessing Intelij was confused here since I usually let it figure out my imports. 
Will this import have real effects or is it just a house keeping thing?

jeremyp-bq: Just house keeping.

Although the wildcard can cause issues if more than one package contains the same class name, and it makes the code harder to read since you don't know which package a class comes from without an IDE.  My preference is to always explicitly list each class, but I think most IDEs by default will fold it into a wildcard if the list grows too large.  Not sure why your IDE is adding an explicit class when the wildcard is already there.

connector/src/main/scala/com/vertica/spark/util/schema/SchemaTools.scala
Aryex: updated to `checkColumnIsComplexType` with comments added.

PR #374, opened 2022-04-19
Update scheduled actions

### LINKED ISSUE #365: Update CI to test against latest patch for each minor version of Spark 3
Currently, existing CI tests to are against the latest version of spark for 3.0.x, 3.1.x, and 3.2.x . We would want to have them test against the latest patch and use the matching version of hadoop.

Aryex:
### Summary
Main workflow will now run against latest Spark minor releases

Added scheduled tests for Vertica 10. It will run weekly each Monday early morning. Note that this will fail until the functional test parameterization is merged in.

### Related Issue

Closes #365 

### Additional Reviewers

@alexey-temnikov 
@alexr-bq 
@jeremyp-bq 
@jonathanl-bq

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -84,7 +84,7 @@ jobs:
       - name: Copy version.properties file from client container
         run: docker exec docker_client_1 cp -r /spark-connector/version.properties /home
       - name: Run the integration tests on Spark 3.0
-        run: docker exec -w /home/functional-tests docker_client_1 sbt run -DsparkVersion="[3.0.0, 3.1.0)" -DhadoopVersion="2.4.0"
+        run: docker exec -w /home/functional-tests docker_client_1 sbt run -DsparkVersion="[3.0.0, 3.1.0)" -DhadoopVersion="[2.7.0, 2.8.0)"
       - name: Remove docker containers
         run: cd docker && docker-compose down
   run-integration-tests-spark-3-1-latest:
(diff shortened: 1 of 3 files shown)


Aryex:
Seems like we will need to update the required checks since this PR changes the action names.

alexr-bq (approved):

.github/workflows/nightly.yml
```diff
@@ -2,8 +2,8 @@ name: Scheduled Test
 
 on:
   schedule:
-    # Every day at 4:18 am
-    - cron: '18 4 * * *'
+    # Every monday at 11:20 GMT
+    - cron: '20 11 * * 1'
```
jeremyp-bq, line 6: Should we rename this file weekly or something related to v10?

Perhaps we should also keep the previous nightly as it can alert us to issues when there is not a recent PR (e.g. if a dependency makes a breaking change).

.github/workflows/main.yml
```diff
@@ -84,10 +84,10 @@ jobs:
       - name: Copy version.properties file from client container
         run: docker exec docker_client_1 cp -r /spark-connector/version.properties /home
       - name: Run the integration tests on Spark 3.0
-        run: docker exec -w /home/functional-tests docker_client_1 sbt run -DsparkVersion="3.0.2" -DhadoopVersion="2.4.0" 
+        run: docker exec -w /home/functional-tests docker_client_1 sbt run -DsparkVersion="[3.0.0, 3.1.0)" -DhadoopVersion="2.4.0"
```
jeremyp-bq, line 87: Regarding the Hadoop version, does anyone know why we use 2.4.0 with Spark 3.0.x, 3.3.1 with 3.1.x, and 3.0.1 with 3.2.x?  Seems a bit strange using an older version of Hadoop with the newer version of Spark.

Looking at the [Spark download page](https://spark.apache.org/downloads.html) and the versions of Hadoop they pair with Spark, I would think the following would be a good match for testing:
* Spark 3.0.x, Hadoop 2.7.x
* Spark 3.1.x, Hadoop 3.2.x
* Spark 3.2.x, Hadoop 3.3.x

.github/workflows/nightly.yml
Aryex: updated.
Added a new action that compiles nightly and only report on failure

.github/workflows/main.yml
Aryex: Updated.

Aryex: I'm not aware of any reason for our hadoop to be of that specific version.

PR #378, opened 2022-04-26
Add partial support for GCS as done for AWS S3

doubleverify-measurement:
### Summary

Current connector support only S3 as staging filesystem we need to use it with databricks on GCP 

### Description

Add some GCS configuration to support the `ALTER SESSION SET GCSAuth=<accessKey>:<secretKey>` in vertica
The access key and secret key are the same as supported for S3

To set you accessKey and secretKey:
``` 
val opt = Map("host" -> host, "table" -> table, "db" -> db, "num_partitions" -> part, "user" -> user, "password" -> password, "staging_fs_url" -> staging_fs_url, "port" -> port, "dbschema" -> dbschema, "gcs_access_key_id" -> "xxx", "gcs_secret_access_key" -> "xxx") 
```

### Related Issue

This code is actually working for us in Databricks GCP using databricks runtime 10.4LTS with spark 3.2.2

### DIFF
--- a/connector/src/main/scala/com/vertica/spark/datasource/core/DSConfigSetup.scala
+++ b/connector/src/main/scala/com/vertica/spark/datasource/core/DSConfigSetup.scala
@@ -30,14 +30,14 @@ import org.apache.spark.sql.SparkSession
 
 
 /**
-  * Interface for taking input of user selected options, performing any setup steps required, and returning the proper configuration structure for the operation.
-  */
+ * Interface for taking input of user selected options, performing any setup steps required, and returning the proper configuration structure for the operation.
+ */
 trait DSConfigSetupInterface[T] {
   /**
-    * Validates and returns the configuration structure for the specific read/write operation.
-    *
-    * @return Will return an error if validation of the user options failed, otherwise will return the configuration structure expected by the writer/reader.
-    */
+   * Validates and returns the configuration structure for the specific read/write operation.
+   *
+   * @return Will return an error if validation of the user options failed, otherwise will return the configuration structure expected by the writer/reader.
+   */
   def validateAndGetConfig(config: Map[String, String]): DSConfigSetupUtils.ValidationResult[T]
 
   /**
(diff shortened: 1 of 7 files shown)


connector/src/main/scala/com/vertica/spark/datasource/core/DSConfigSetup.scala
```diff
@@ -299,17 +335,17 @@ object DSConfigSetupUtils {
   }
 
   private def getAWSArg(visibility: Visibility)(
```
jeremyp-bq, line 337: We should rename some of these functions since they are not AWS-specific (`getAWSArg`, `getAWSArgFromConnectorOption`, `getAWSArgFromSparkConfig`, and `getAWSArgFromEnvVar` - just drop the `AWS`?).

connector/src/main/scala/com/vertica/spark/datasource/core/DSConfigSetup.scala
```diff
@@ -219,6 +233,28 @@ object DSConfigSetupUtils {
...
   }
 
+
+  def getGCSAuth(config: Map[String, String]): ValidationResult[Option[GCSAuth]] = {
+    val visibility = Secret
+    val accessKeyIdOpt = getAWSArg(visibility)(
+      config,
+      "gcs_access_key_id",
+      "",
+      "").sequence
```
jeremyp-bq, line 243: I think we should have some fallbacks, like with AWS args.  In this case the Spark arg `google.cloud.auth.client.id` and the env var `GOOGLE_CLIENT_ID` (I can't find documentation on an official Google env var name).  And for the secret below, `google.cloud.auth.client.secret` and `GOOGLE_CLIENT_SECRET`.

doubleverify-measurement: There is no official environment environment variable like for AWS this is why I've left them empty. However we can still use env variable

PR #396, opened 2022-05-16
weekly-v10-test-fix

Aryex:
Test no longer print version from Vertica. 

### Additional Reviewers
@alexr-bq 
@jeremyp-bq 
@jonathanl-bq

### DIFF
--- a/.github/workflows/weekly-v10-test.yml
+++ b/.github/workflows/weekly-v10-test.yml
@@ -29,8 +29,6 @@ jobs:
         uses: actions/checkout@v2
       - name: Run docker compose
         run: cd docker && docker-compose up -d
-      - name: Print Vertica version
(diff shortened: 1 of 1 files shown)


.github/workflows/weekly-v10-test.yml
```diff
@@ -29,8 +29,6 @@ jobs:
         uses: actions/checkout@v2
       - name: Run docker compose
         run: cd docker && docker-compose up -d
-      - name: Print Vertica version
-        run: docker exec docker_vertica_1 vsql -c "select version();"
```
jeremyp-bq, line 33: Note that the other times we use vsql in GitHub actions we use the nick-invision/retry@v2 plugin so that we can retry while the container is still starting and possibly not available (in case we need to perform other vsql commands in the future).

jeremyp-bq (approved):

PR #400, opened 2022-05-19
Fixing multiple JDBC connections

### LINKED ISSUE #398: [BUG] Connector write/read opens multiple connections
## Problem Description
It has been reported that the connector is creating multiple jdbc sessions. The connector should only create a constant number of JDBC connections on the driver node to prepare for read/write. The number of connections should not scale with executors created. 

---

## Environment
- Spark version: N/A
- Hadoop version: N/A
- Vertica version: N/A
- Java version: N/A
- Additional Environment Information: pyspark on Jupyter client

Aryex:
### Summary

Fixed an issue with the connector creating multiple JDBC connections on write. 

### Description

When checking for jdbc layer is closed, included a lazy initialization check to ensure we don't initialize jdbc connection when it wasn't used. 
This fix is not ideal and in the future, we should look refactor executor codes so that they do not hold a jdbc layer reference as they do not need to use jdbc.

#### Other changes
 * Post merge functional tests added
 * Docker setup to expose Spark UI
 * Fixed weekly tests hadoop version

### Related Issue

Closes #398 

### Additional Reviewers

@alexr-bq 
@jeremyp-bq 
@jonathanl-bq


connector/src/main/scala/com/vertica/spark/datasource/jdbc/VerticaJdbcLayer.scala
```diff
@@ -179,24 +179,31 @@ class VerticaJdbcLayer(cfg: JDBCConfig) extends JdbcLayerInterface {
...
 
   private def createClientLabel: String = {
     val authMethod = this.cfg.auth match {
       case BasicJdbcAuth(_, _) => "-p"
       case KerberosAuth(_, _, _, _) => "-k"
     }
-    val sparkVersion = SparkSession.active.sparkContext.version
+    val sparkVersion = SparkSession.getActiveSession match {
+      case Some(session) => session.sparkContext.version
+      case None => "0.no-active-session"
```
jeremyp-bq, line 205: What's the significance of this value, why not "0.0" or "na" if we can't get the version?

docker/docker-compose.yml
```diff
@@ -9,6 +9,8 @@ services:
     stdin_open: true
     ports:
       - "5005:5005"
+      - "8080:8080"
+      - "8081:8081"
```
jeremyp-bq, line 13: Why do we need to map these ports?

functional-tests/default-config.sh
```diff
@@ -0,0 +1,10 @@
+echo -e 'functional-tests {
```
jeremyp-bq, line 1: What is this for, was this for testing possible issues with quoted values?

functional-tests/src/main/scala/Main.scala
```diff
@@ -291,8 +296,7 @@ object Main extends App {
     println(s"${result.suiteName} $status")
     if (testFailed) {
       result.testsFailed.foreach(failedTest => {
-        println(s" - FAILED: ${failedTest.testName}, message:")
-        println(s"  ${failedTest.message}")
```
jeremyp-bq, line 295: Are you removing the error message since it is in the output above (i.e. this part is only a summary)?  Might still be useful to print out the root cause message at the end of the test results.

functional-tests/src/main/scala/com/vertica/spark/functests/endtoend/RemoteTests.scala
```diff
@@ -0,0 +1,58 @@
+package com.vertica.spark.functests.endtoend
+
+import com.vertica.spark.config.{FileStoreConfig, JDBCConfig}
+import org.apache.spark.sql.{Row, SaveMode, SparkSession}
+import org.apache.spark.sql.types.{IntegerType, StructField, StructType}
+
+/**
+ * Test suites for submitting to a remote driver.
```
jeremyp-bq, line 8: What is the difference between the "remote" tests and other tests, is it just that this one is meant to be run with the sh script which starts up a Spark master and worker?  Maybe just add that to the comment here so people know what the remote driver/tests means.

connector/src/main/scala/com/vertica/spark/datasource/jdbc/VerticaJdbcLayer.scala
Aryex: Fixed.

docker/docker-compose.yml
Aryex: They are for Spark UI. 8080 is for the driver ui and 8081 is for the executor node.

functional-tests/default-config.sh
Aryex: Just a convenient script to reset the config

functional-tests/src/main/scala/Main.scala
Aryex: That was my initial intention. However, I found that the error message would sometime contain the stack trace which clutters the output.

connector/src/main/scala/com/vertica/spark/datasource/core/factory/VerticaPipeFactory.scala
```diff
@@ -90,36 +94,39 @@ object VerticaPipeFactory extends VerticaPipeFactoryInterface {
...
 
+  //scalastyle:off
   override def getWritePipe(config: WriteConfig): VerticaPipeInterface with VerticaPipeWriteInterface = {
+    val thread = Thread.currentThread.getName + ": "
+    logger.debug(thread + "Getting write pipe")
     config match {
       case cfg: DistributedFilesystemWriteConfig =>
         writeLayerJdbc = checkJdbcLayer(writeLayerJdbc, cfg.jdbcConfig)
-        val verticaVersion = VerticaVersionUtils.getVersion(writeLayerJdbc.get)
+        val jdbcLayer = writeLayerJdbc.orNull
```
jonathanl-bq, line 104: Can we avoid the usage of null here? This is kind of sneaky and it'll make things harder to reason with if it ends up being null here.

jonathanl-bq (changes requested):

functional-tests/src/main/scala/com/vertica/spark/functests/endtoend/RemoteTests.scala
Aryex: Updated. Currently, all of our tests initialize a spark session with `master("local[*]")` which essentially hardcoded them to run on local spark threads, even if we submit them. The RemoteTests overrode it so we can specify the master URL when submitting them.

connector/src/main/scala/com/vertica/spark/datasource/core/DSWriter.scala
```diff
@@ -52,7 +53,10 @@ trait DSWriterInterface {
  * @param uniqueId Unique identifier for this specific writer. The writer for each partition should have a different ID.
  * @param pipeFactory Factory returning the underlying implementation of a pipe between us and Vertica, to use for write.
  */
-class DSWriter(config: WriteConfig, uniqueId: String, pipeFactory: VerticaPipeFactoryInterface = VerticaPipeFactory) extends DSWriterInterface {
+class DSWriter(config: WriteConfig, uniqueId: String, pipeFactory: VerticaPipeFactoryInterface = VerticaPipeFactory, jdbc: Boolean = false) extends DSWriterInterface {
```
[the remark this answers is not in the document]
Aryex: Removed.

connector/src/main/scala/com/vertica/spark/datasource/core/factory/VerticaPipeFactory.scala
Aryex: Yeah, its not ideal but checkJdbcLayer() should guarantee that we never have a null jdbc reference in the optional.

docker/docker-compose.yml
jeremyp-bq: I think we should add them to the kerberos docker-compose file as well then to keep it consistent.

I could see us creating a multi-node Spark cluster within Docker at some point in the future rather than creating a single master and worker in the client container.  But that is down the road.

Aryex: updated

connector/src/main/scala/com/vertica/spark/datasource/core/factory/VerticaPipeFactory.scala
Aryex: updated

connector/src/main/scala/com/vertica/spark/datasource/core/DSWriter.scala
```diff
@@ -52,7 +53,10 @@ trait DSWriterInterface {
...
+  private val logger: Logger = LogProvider.getLogger(classOf[DSWriter])
+  private val thread = Thread.currentThread().getName + ": "
+  logger.debug(thread + "Initializing writer")
```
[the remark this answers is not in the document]
Aryex: added a ticket for it

alexr-bq:
My only additional thought here is - would it be possible to modify the jdbc configuration so that it is never sent down to the executors at all, making it impossible they'll ever try a proper jdbc connection?

Aryex:
> My only additional thought here is - would it be possible to modify the jdbc configuration so that it is never sent down to the executors at all, making it impossible they'll ever try a proper jdbc connection?

It's possible, but we will need to refactor the pipes which will have a larger scope than this.

jonathanl-bq (approved):

jeremyp-bq (approved):

PR #408, opened 2022-05-27
Release 3.2.2

Aryex:
### Summary

* Version number updated for release 3.2.2. Note that in previous PRs, GitHub actions were updated to use wildcard so we no longer need to update their version string.
* Added a fix for read to not create multiple JDBC connections. 

### Additional Reviewers
@alexey-temnikov 
@alexr-bq 
@jeremyp-bq 
@jonathanl-bq

### DIFF
--- a/functional-tests/src/main/scala/com/vertica/spark/functests/endtoend/RemoteTests.scala
+++ b/functional-tests/src/main/scala/com/vertica/spark/functests/endtoend/RemoteTests.scala
@@ -14,6 +14,7 @@ class RemoteTests(readOpts: Map[String, String], writeOpts: Map[String, String],
   // This suite uses a spark session without a master specified, giving control to the submitter.
   override lazy val spark: SparkSession = SparkSession.builder()
     .appName("Vertica Connector Test Prototype")
+    .master("local[4]")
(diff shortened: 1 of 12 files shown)


jeremyp-bq (approved):

functional-tests/src/main/scala/com/vertica/spark/functests/endtoend/RemoteTests.scala
```diff
@@ -45,14 +46,30 @@ class RemoteTests(readOpts: Map[String, String], writeOpts: Map[String, String],
...
+          ("num_partitions"-> "30") +
+          ("max_row_group_size_export_mb" -> "1") +
+          ("max_file_size_export_mb" -> "1"))
+        .load()
+        .write.format("json").mode(SaveMode.Overwrite).save("./test-output")
+
+      rs = stmt.executeQuery(query)
+      assert(rs.next)
+      val sessionCountRead = rs.getLong(1)
+      assert(sessionCountRead == initialJdbcSessionCount + 4)
```
alexr-bq, line 65: Why is read creating 4 connections?

Aryex: Read was done after a write, which also created 2 connections. Thus, we + 4 onto the initial connection count.

PR #410, opened 2022-05-30
Handle query with complex type

### LINKED ISSUE #405: [BUG] Query connector parameter fails during table introspection
## Environment
- Spark version: 3.2
- Hadoop version: 3.3.1
- Vertica version: 10.1
- Java version: openjdk 1.8.0_312

## Problem Description

The documentation states that is possible to load data from Vertica into Spark either by using the 'table' option or the 'query' option. However, when using the 'query' option I get an error that points to the table_name not being correctly parsed from the 'query' string:

Py4JJavaError: An error occurred while calling o194.showString.
: com.vertica.spark.util.error.ConnectorException: Query result is empty 
 QUERY:[SELECT data_type_id FROM columns WHERE table_name='' AND column_name='AREANAME'] 

(full stack on attached file)

I believe the connector is correctly getting all the columns in the table defined on the query, then it uses the v_catalog.columns system table to get the data_type_id for each column. However, on the generated query the table_name is left empty.

Setting the 'table' parameter does not solve the issue (by itself it works, with 'query' as well, it fails)

Below is a PySpark snippet that causes the problem:

connOptions = {
    "host" : ""####",,
    "user" : "dbadmin",
    "password" : ""####",",
    "db" : 'MYDB',
    "dbschema" : 'MYSCHEMA',
    'aws_enable_ssl' : False,
    "staging_fs_url" : "s3a://spark-staging/vertica",
    "aws_endpoint" : "####",
    "aws_access_key_id" : "####",
    "aws_secret_access_key" : "#####",
    "aws_region" : "us-east-1",
    "aws_enable_path_style" : True,
    "query" : "SELECT * FROM AREAS"
}

table = spark.read \
  .format("com.vertica.spark.datasource.VerticaSource") \
  .options(**connOptions) \
  .load()

table.show()

---
[callstack.txt](https://github.com/vertica/spark-connector/files/8782429/callstack.txt)

Aryex:
### Summary

The connector cannot handle queries specified by the option `query` that return complex types. This is because a query's schema is not recorded in Vertica system tables, which is what we are using to infer the structure of complex type columns. 

Thus, we add are now checking through JDBC type codes and returning a meaningful message. Note that for Vertica 10, we cannot check because complex types are reported as strings through JDBC.

### Related Issue

Closes #405 

### Additional Reviewers

@alexey-temnikov 
@alexr-bq 
@jeremyp-bq 
@jonathanl-bq


connector/src/main/scala/com/vertica/spark/util/schema/SchemaTools.scala
```diff
@@ -264,23 +264,13 @@ class SchemaTools extends SchemaToolsInterface {
...
-    case TableQuery(query, _) =>
-      ColumnInfoQueryData("", "" , "SELECT * FROM (" + query + ") AS x WHERE 1=0")
-  }
-
   def getColumnInfo(jdbcLayer: JdbcLayerInterface, tableSource: TableSource): ConnectorResult[Seq[ColumnDef]] = {
-    val tableInfo = getColumnInfoQueryData(tableSource)
-    jdbcLayer.query(tableInfo.emptyQuery) match {
+    val emptyQuery = tableSource match {
+      case tb: TableName => "SELECT * FROM " + tb.getFullTableName + " WHERE 1=0"
+      case TableQuery(query, _) => "SELECT * FROM (" + query + ") AS x WHERE 1=0"
```
jeremyp-bq, line 270: Not needed for this PR, but I'm wondering if `LIMIT 0` would be faster than `WHERE 1=0` (`LIMIT 0` is certainly easier to read).

jeremyp-bq, line 270: Not needed for this PR, but I'm wondering if `LIMIT 0` would be faster than `WHERE 1=0` (`LIMIT 0` is certainly easier to read).

jeremyp-bq (approved):

connector/src/main/scala/com/vertica/spark/util/schema/SchemaTools.scala
Aryex: I think it depends on how Vertica optimizes the query? since the condition is constant it might be able to pick it up and reduce it to false, returning an empty query.

PR #412, opened 2022-05-30
Fix issues with execution of remote tests

jeremyp-bq:
### Summary

Was unable to run the tests with the default 1g of driver memory, and the test failed since it tried to use a closed statement.

### Additional Reviewers

@alexey-temnikov
@alexr-bq
@Aryex
@jonathanl-bq

### DIFF
--- a/functional-tests/submit-functional-tests.sh
+++ b/functional-tests/submit-functional-tests.sh
@@ -5,4 +5,4 @@ export SPARK_HOME=/opt/spark
 export PATH=$PATH:$SPARK_HOME/bin:$SPARK_HOME/sbin
 start-master.sh -h localhost
 start-worker.sh -h localhost spark://localhost:7077
-spark-submit --master spark://localhost:7077 target/scala-2.12/spark-vertica-connector-functional-tests-assembly-$CONNECTOR_VERSION.jar $1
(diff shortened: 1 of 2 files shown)


Aryex (approved):

functional-tests/submit-functional-tests.sh
```diff
@@ -5,4 +5,4 @@ export SPARK_HOME=/opt/spark
 export PATH=$PATH:$SPARK_HOME/bin:$SPARK_HOME/sbin
 start-master.sh -h localhost
 start-worker.sh -h localhost spark://localhost:7077
-spark-submit --master spark://localhost:7077 target/scala-2.12/spark-vertica-connector-functional-tests-assembly-$CONNECTOR_VERSION.jar $1
+spark-submit --master spark://localhost:7077 --driver-memory 2g target/scala-2.12/spark-vertica-connector-functional-tests-assembly-$CONNECTOR_VERSION.jar $1
```
Aryex, line 8: Do you know how this would be handled on github action? would it just give us more memory?

jeremyp-bq: Since we don't host our own runners, we use the provided runners (looks to be 7GB RAM, which should be able to handle upping the Spark memory from the default of 1GB to 2GB):
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

PR #414, opened 2022-06-01
System table queries refactor

Aryex:
### Summary

System tables are used by schema tools to infer complex data types structures. This PR refactored the existing code, abstracting the queries into the abstract class VerticaTables, and its concrete classes. 

VerticaTable purpose is to handle querying and collecting row data from a table. Specific information about a table, like a table name, columns to query, and how to build a row, are provided by its subclasses. 

When reviewing, VerticaTables and its subclasses are good places to start. Afterwards, you can see how they are used in schema tools.

### Changes
* Added abstract VerticaTable class containing base implementation for table queries.
* Added ColumnsTable, TypesTable, and ComplexTypesTable extended from VerticaTable for querying with those system tables.
* SchemaTools and SchemaToolsV10 updated. 
  * These classes were the ones making queries to Vertica System tables.
  * Direct JDBC queries are replaced with the table abstractions.
* Moved code related to getting complex data type from Vertica to ComplexTypeSchemaSupport to improve readability.
* Added support to get precision and scale for numeric types
* Added unit tests for VerticaTable and its concrete implementations.
* Updated other unit tests.

### Additional Reviewers
@alexey-temnikov 
@alexr-bq 
@jeremyp-bq 
@jonathanl-bq


alexr-bq:
Just feedback on PR description:
"SchemaTools and SchemaToolsV10 updated" - how?
"Purely to fix scalastyle warning for method count." - I don't think this is necessarily a sufficient reason. If we disagree with the warning we can ignore it. If not, would be more appropriate to say "to reduce complexity" or "to improve readability" or something like that.

connector/src/main/scala/com/vertica/spark/util/query/ComplexTypesTable.scala
```diff
@@ -0,0 +1,53 @@
...
+// scalastyle:off magic.number
+class ComplexTypesTable(jdbcLayer: JdbcLayerInterface)
+  extends VerticaTable[ComplexTypeInfo](jdbc = jdbcLayer) {
+
+  override protected def tableName: String = "complex_types"
+
+  override protected def columns: Seq[String] = List("type_id", "type_name", "field_id", "field_type_name", "numeric_scale", "type_kind", "numeric_precision")
+
+  override protected def buildRow(rs: ResultSet): ComplexTypeInfo =
+    ComplexTypeInfo(
```
alexr-bq, line 34: I'm confused what's going on here

Aryex: When a query is made to complex_types table, the ResultSet is given to this function to construct an object holding some row data (ComplexTypeInfo in this case)

Aryex: If you take a look at VerticaTable.querySelectWhere, you'll see that it is constructing a query, then calling this function for each result set.

connector/src/main/scala/com/vertica/spark/util/error/ErrorHandling.scala
```diff
@@ -580,4 +589,22 @@ case class HDFSConfigError() extends ConnectorError {
 case class JobAbortedError() extends ConnectorError {
   def getFullContext: String = "Writing job aborted. Check spark worker log for specific error."
 }
+case class VerticaColumnNotFound(colName: String, tableName: String, schema: String) extends ConnectorError {
```
jeremyp-bq, line 592: This error only seems to be imported, but never used.

connector/src/main/scala/com/vertica/spark/util/error/ErrorHandling.scala
```diff
@@ -86,6 +86,9 @@ case class SchemaDiscoveryError() extends ConnectorError {
     "There may be an issue with connectivity to the database."
 
 }
+case class UndefinedError() extends ConnectorError {
```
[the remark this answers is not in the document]
Aryex: It was intended to use in situations where a `ConnectorError` is not yet decided but is required as part of a function return. I'm not sure something like this should be added.

connector/src/main/scala/com/vertica/spark/util/error/ErrorHandling.scala
Aryex: removed.

connector/src/main/scala/com/vertica/spark/util/error/ErrorHandling.scala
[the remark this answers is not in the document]
jeremyp-bq: I don't think it is needed, we can add specific errors as they are needed.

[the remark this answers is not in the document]
Aryex: removed

PR #420, opened 2022-06-07
Json Protoype

### LINKED ISSUE #370: [SPIKE] JSON export option

Aryex:
### Summary

Contains code demonstrating how Spark JsonTable would be used to handle reading JSON. Refer to #370 for an operation flow diagram. In general, the code aims to wrap Spark's Json implementation to reuse its functionalities.

As is, this PR allows the connector to read Json data without cleanup after each read. Note that we still do a general deletion on application ends

### Related Issue

Close #370 

### Additional Reviewers
@alexey-temnikov 
@alexr-bq 
@jeremyp-bq 
@jonathanl-bq


connector/src/main/scala/com/vertica/spark/config/ReadConfig.scala
```diff
@@ -126,6 +132,12 @@ final case class DistributedFilesystemReadConfig(
     this.aggPushedDown = pushdownAgg
   }
 
+  override def setUseJson(use: Boolean): Unit = {
+    this.jsonRead = use
```
jeremyp-bq, line 136: Why do we need a jsonRead variable when there is already a json variable (this function works on a different variable than the useJson function)?

Aryex: mb. It's a duplicate option. Should be removed now.

jeremyp-bq:
Should we update the README and add tests/examples, or will that be covered in a later ticket when JSON support has been integrated with complex types?

Aryex:
> Should we update the README and add functional tests/examples, or will that be covered in a later ticket when JSON support has been integrated with complex types?

It would be covered in a separate PR.

connector/src/main/scala/com/vertica/spark/datasource/wrappers/VerticaScanWrapper.scala
```diff
@@ -0,0 +1,52 @@
...
+   * Calls the wrapped scan to plan inputs. Then process them into [[VerticaFilePartition]] with partitioning info
+   * */
+  override def planInputPartitions(): Array[InputPartition] = {
+    val partitioningRecords = scala.collection.mutable.Map[String, Int]()
+
+    def recordPartitionedFile(file: PartitionedFile) = {
+      val key = file.filePath
+      val count = partitioningRecords.getOrElse(key, 0)
+      partitioningRecords.put(key, count + 1)
+      VerticaPartitionedFile(file, count)
```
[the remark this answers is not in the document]
Aryex: The id is expected to start from 0, so we would want the new id to = count - 1, or in this case just count before being + 1

jeremyp-bq (approved):

Issue #416, opened 2022-06-03
[Json Read Support] Cleanup


Aryex:
## Description
Currently, the Json read prototype does not perform any cleanup. This will need to be implemented to be able to cleanup the staging area.

jeremyp-bq:
See also #181, and consider error cases if possible.

alexey-temnikov:
Initial PR is in review - https://github.com/vertica/spark-connector/pull/427.

PR #426, opened 2022-06-15
Read pipe cleanup refactor

Aryex:
### Summary

* Extracted read pipe cleanup logic into a separate class. It is intended to be re-used by JSON reader. 
* Added new supporting interfaces. 
* Simplified existing partition interfaces:
  * Previously, ParquetFileRange and VerticaDistributedFileSystem define fields which are not needed to be optional. They are refactored to normal fields. Related files are updated.

### Related Issue

#416 

### Additional Reviewers
@alexey-temnikov 
@alexr-bq 
@jeremyp-bq 
@jonathanl-bq

### DIFF
--- a/connector/src/main/scala/com/vertica/spark/datasource/partitions/PartitionCleanup.scala
+++ b/connector/src/main/scala/com/vertica/spark/datasource/partitions/PartitionCleanup.scala
@@ -15,8 +15,18 @@ package com.vertica.spark.datasource.partitions
 
 import org.apache.spark.sql.connector.read.InputPartition
 
-trait PartitionCleanup extends InputPartition {
-  def getCleanUps: Seq[Cleanup]
+/**
+ * Mixin trait for [[InputPartition]] that contains partitions cleanup information
+ * */
+trait PartitionCleanup {
 
+  /**
+   * @return returns a list of partition cleanup information
+   * */
+  def getCleanupInformation: Seq[Cleanup]
+
+  /**
+   * @return return a mapping of filename to their partition count
+   * */
   def getPartitioningRecord: Map[String, Int]
 }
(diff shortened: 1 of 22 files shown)


jonathanl-bq (commented):
I don't have time to do a really thorough review, but I think we should avoid exposing index math in our traits. It feels like a lower level implementation detail more than something we should expose in a trait, but I don't know if there's a better justification for this current design.

Aryex:
> I don't have time to do a really thorough review, but I think we should avoid exposing index math in our traits. It feels like a lower level implementation detail more than something we should expose in a trait, but I don't know if there's a better justification for this current design.

Updated. Also changed the names of the traits to Cleanup and PartitionCleanup

connector/src/main/scala/com/vertica/spark/datasource/partitions/PartitionCleanup.scala
```diff
@@ -0,0 +1,22 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.vertica.spark.datasource.partitions
+
+import org.apache.spark.sql.connector.read.InputPartition
+
+trait PartitionCleanup extends InputPartition {
+  def getCleanUps: Seq[Cleanup]
```
jeremyp-bq, line 19: Minor nit: getCleanUps when we use Cleanup (up not capitalized) everywhere else.

Aryex: Updated

Aryex:
Updated trait name to better reflect the implementor capabilities
Cleanup -> PortionId: Implementors can identify themselves amongst other portions from the same file.
PartitionCleanup -> Cleanup: Implementors contains cleanup information.

jeremyp-bq (approved):

PR #432, opened 2022-06-20
Fix failing v10 functional tests

Aryex:
### Summary
Fix an issue with the depth attribute not present in v10 metadata

### Reviewers
@alexey-temnikov 
@jeremyp-bq 
@jonathanl-bq


connector/src/main/scala/com/vertica/spark/util/schema/SchemaTools.scala
```diff
@@ -715,13 +715,19 @@ class SchemaToolsV10 extends SchemaTools {
       case Left(value) => Left(value)
       case Right(columnInfo: ColumnInfo) =>
         val verticaType = columnInfo.verticaType
+
+        // Native array case
         if(verticaType > VERTICA_NATIVE_ARRAY_BASE_ID && verticaType < VERTICA_SET_MAX_ID) {
-          val dummyChild = ColumnDef("", java.sql.Types.VARCHAR, "STRING", 0, 0, false, false, Metadata.empty)
+          val metadata = new MetadataBuilder().putLong(MetadataKey.DEPTH, 0).build()
```
jeremyp-bq, line 721: Could move this metadata and the one on line 727 outside the if/else since it is the same for both cases.

jeremyp-bq (approved):

PR #448, opened 2022-06-30
Support query option schema

### LINKED ISSUE #437: [BUG] Loading data from vertical using the 'query' option always uses the default schema
## Environment
- Spark version: 3.2
- Hadoop version: 3.3.1
- Vertica version: 10.1
- Java version:
- Additional Environment Information:

---

## Problem Description
When loading vertica data into a dataframe using the 'query' option, the 'dbschema' seems to be ignored. Prefixing the SQL tables with the schema works as expected, but this is not always convenient. In my case, I am getting the SQL from a semantic layer, which doesn't add the scheme to the table names and expects the correct schema to be configured as default.

1. Steps to reproduce:

Use spark.read.format, and define the 'query' and 'dbschema' options. Set 'dbschema' to a schema other than the default. Use a SQL query that doesn't prepend schema to the tablenames and that targets a table that only exists on dbschema. The call incorrectly fails.

2. Expected behaviour:

The 'dbschema' option should be honored when 'query' is used

3. Actual behaviour:

'dbschema' is not honored

5. Error message/stack trace:
6. Code sample or example on how to reproduce the issue:

---

## Spark Connector Logs
- Add related logs entries here.

Aryex:
### Summary

Adding support for using `query` option with `dbschema` option.

### Description

With both defined, the schema specified in `dbschema` will be appended to the table name in the query's FROM clause. This will also work when string literals are used.
The `dbschema` _will not be appended_ if:

1. The FROM clause is a query.
2. The FROM clause already has a schema defined.

For reviewing, main code logic starts at `SchemaTools.addDbSchemaToQuery()`.

### Related Issue

Close #437 

### Additional Reviewers

@jeremyp-bq 
@jonathanl-bq 
@alexey-temnikov 
@alexr-bq

### DIFF
--- a/connector/src/main/scala/com/vertica/spark/util/schema/SchemaTools.scala
+++ b/connector/src/main/scala/com/vertica/spark/util/schema/SchemaTools.scala
@@ -142,6 +142,16 @@ trait SchemaToolsInterface {
    * @param schema schema of the table
    */
   def checkValidTableSchema(schema: StructType): ConnectorResult[Unit]
+
+  /**
+   * Given a query, append the schema to FROM clause table.
+   * Ignore if FROM clause is a sub query or already has a schema
+   *
+   * @param query an SQL query
+   * @param schema an optional schema to be added.
+   * @return the query with the added schema.
+   * */
+  def addDbSchemaToQuery(query: String, schema: Option[String]): String
 }
 
 class SchemaTools(ctTools: ComplexTypesSchemaTools = new ComplexTypesSchemaTools) extends SchemaToolsInterface {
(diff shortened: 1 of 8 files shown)


connector/src/main/scala/com/vertica/spark/util/schema/SchemaTools.scala
```diff
@@ -682,6 +692,89 @@ class SchemaTools(ctTools: ComplexTypesSchemaTools = new ComplexTypesSchemaTools
     // Use recursion to break early
     findEmptyColumnName(schema.fields.toList)
   }
+
+  private def noDotOutsideStringLiteral(str: String): Boolean = {
```
jeremyp-bq, line 696: Should add a scaladoc to a few functions here.  Also, can we improve the naming of a few of these functions?

For example:
* noDotOutsideStringLiteral -> containsSchema
* dotOutsideStringLiteral -> containsSchemaHelper
* recursion -> addDbSchemaToQueryHelper

Aryex: Updated. Regex is used instead to replace the noDotOutsideStringLiteral(). Functional names are also updated.

Aryex: about the scaladoc, do you mean for `addDbSchemaToQuery`? It is defined at the trait definition.

jeremyp-bq: That's sufficient.

jeremyp-bq (approved):

PR #457, opened 2022-07-07
Updating Automated Tests

### LINKED ISSUE #402: Document Github CI
Since we have quite a few GitHub actions, we should document what they do in Contributing.md

### LINKED ISSUE #421: Cleanup/organization of GitHub actions and tests.
## Description 
Review Github actions and see which one is needed and not.

### LINKED ISSUE #463: [BUG] When reading with JSON, staging files are not deleted if tables are empty.
## Environment
- Vertica Spark Connector version: 3.3.0
---

## Problem Description
- When reading from an empty table using option `json`, connector fails if the table is empty.
---

Aryex:
### Summary

This PR contains updates to the automated testing. In general, each PR now only runs unit tests, scalastyles, and functional test against latest Spark. Regression testing and other tests are done nightly.

### Description
Replacing `main.yml` with `on-pull-request.yml` workflow, now run unit tests, scalastyle checks, and functional tests against the latest version of Spark.

Updated `nightly.yml` workflow to perform long-running tests. This includes:

- Regression testing against different Spark 3.x vs Vertica 10, 11, and 12
- Testing against GCS and S3
- Testing JSON option against Spark 3.x vs Vertica 12. This involves updating a few functional tests to handle json option.

Functional tests updated:
- Added an exclude option, allowing the user to exclude test suites from a run.
- Modified checks for a few tests in ComplexTypeTests and EndToEndTests to account for when using JSON option.
- `build.sbt` no longer specify `org.apache.hadoop.hadoop-hdfs` dependency. `spark-core` already defines it. 
- `build.sbt` will now additionally look for `sparkVersion` and `hadoopAwsVersion` from the environment variables. Previously it was only checking in JVM's properties.

Readmes updated 
- A Readme added to `.github/workflows`
- Updated S3 to reflect new learning about setting up aws-hadoop connector
- Updated functional tests readme to reflect changes.

Bug fixes:
- Fixed a bug where if the exported JSON folder was empty, the connector would fail as the underlying Spark JSON reader reports nothing to read. Changes start at `VerticaJsonScan.batch` where after the data was exported and Spark's Json reader has been initialized, we then delete the folder if empty. 

### Related Issue

Close #421
Close #402
Close #463


functional-tests/src/main/scala/com/vertica/spark/functests/endtoend/EndToEndTests.scala
```diff
@@ -3590,7 +3614,7 @@ class EndToEndTests(readOpts: Map[String, String], writeOpts: Map[String, String
     fsLayer.createDir(fsConfig.address, "777")
   }
 
-  it should "use schema metadata to override col size when creating an external table with varchar/varbinary type" in {
+  ignore should "use schema metadata to override col size when creating an external table with varchar/varbinary type" in {
```
jeremyp-bq, line 3617: Maybe add a comment above each of these two ignored tests so we know why they are ignored.

functional-tests/build.sbt
```diff
@@ -25,8 +25,15 @@ name := "spark-vertica-connector-functional-tests"
...
 version := versionProps.value.getProperty("connector-version")
 
-val sparkVersion = Option(System.getProperty("sparkVersion")).getOrElse("3.3.0")
-val hadoopVersion = Option(System.getProperty("hadoopVersion")).getOrElse("3.3.0")
+val sparkVersion = Option(System.getProperty("sparkVersion")) match {
+  case Some(sparkVersion) => sparkVersion
+  case None => sys.env.getOrElse("SPARK_VERSION", "3.3.0")
+}
+
+val hadoopAwsVersion = Option(System.getProperty("HADOOP_AWS_VERSION")) match {
```
jeremyp-bq, line 33: Should this be `hadoopVersion` (or `hadoopAwsVersion`)?

functional-tests/src/main/scala/com/vertica/spark/functests/endtoend/EndToEndTests.scala
```diff
@@ -42,6 +44,11 @@ abstract class EndToEnd(readOpts: Map[String, String], writeOpts: Map[String, St
   protected val fsConfig: FileStoreConfig = FileStoreConfig(readOpts("staging_fs_url"), "", false, fileStoreConfig.awsOptions, fileStoreConfig.gcsOptions)
   protected val fsLayer = new HadoopFileStoreLayer(fsConfig, None)
   protected val VERTICA_SOURCE = "com.vertica.spark.datasource.VerticaSource"
+  protected lazy val verticaVersion: Version = {
+    val rs = conn.createStatement().executeQuery("select version();")
+    rs.next()
+    VerticaVersionUtils.parseVerticaVersionString(rs.getString(1)).getOrElse(VERTICA_DEFAULT)
+  }
```
Aryex, line 51: Should be removed next push. Not used in tests.

connector/src/main/scala/com/vertica/spark/datasource/json/VerticaJsonScan.scala
```diff
@@ -47,7 +49,15 @@ class VerticaJsonScan(config: ReadConfig, readConfigSetup: DSConfigSetupInterfac
...
-          batchFactory.build(partitionInfo.rootPath, Some(readSchema()), jsonReadConfig, sparkSession)
+          val batch = batchFactory.build(partitionInfo.rootPath, Some(readSchema()), jsonReadConfig, sparkSession)
+
+          val files = fsLayer.getFileList(partitionInfo.rootPath).getOrElse(ErrorHandling.logAndThrowError(logger, InitialSetupPartitioningError()))
+          if(files.isEmpty){
+            new CleanupUtils().cleanupAll(fsLayer, partitionInfo.rootPath)
+            batch
+          }else{
+            batch
+          }
```
jeremyp-bq, line 60: Nit: this could be simplified to always return batch after the if statement (which only contains the cleanupAll call).

.github/workflows/on-pull-request.yml
```diff
@@ -19,11 +19,36 @@ jobs:
...
+          fail_ci_if_error: true
+          verbose: true
+  run-integration-tests:
+    runs-on: ubuntu-latest
+    needs: [build]
     env:
-      VERTICA_VERSION: 10.1.1-0
+      VERTICA_VERSION: 12.0.0-0
+      SPARK_VERSION: 3.3.0
+      HADOOP_AWS_VERSION: 3.3.2
```
jeremyp-bq, line 51: Do we want to hard-code the values here, or do we want to fallback on the default (which should be the latest versions we support, provided we have updated our build.sbt files)?

.github/workflows/nightly.yml
```diff
@@ -20,19 +19,326 @@ jobs:
...
           name: build-jar-file
           path: /home/runner/work/spark-connector/spark-connector/connector/target/scala-2.12/spark-vertica-connector_2.12-*.jar
+  run-spark-integration-tests:
+    name: Run regression tests against different versions of Vertica
+    runs-on: ubuntu-latest
+    needs: build
+    strategy:
+      fail-fast: false
+      matrix:
+        spark: [ "[3.0.0, 3.1.0)", "[3.1.0, 3.2.0)", "[3.2.0, 3.3.0)", "[3.3.0, 3.4.0)" ]
```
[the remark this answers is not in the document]
jeremyp-bq: I believe you can use a complex value like this:
`runtime: [ { spark: "[3.0.0, 3.1.0)", hadoop: "2.7.0" }, ... ]`

Then refer to it as `matrix.runtime.spark` or `matrix.runtime.hadoop`.

.github/workflows/nightly.yml
```diff
@@ -20,19 +19,326 @@ jobs:
...
+          path: ./functional-tests/lib/
+      - name: Build the functional-tests project
+        run: cd functional-tests && sbt assembly
+      - name: Build client image
+        run: docker build -t client ./docker
+      - name: Run docker compose
+        run: cd docker && docker-compose up -d
+        env:
+          HADOOP_VERSION: "3.3.0"
+          SPARK_VERSION: "3.2.1"
```
jeremyp-bq, line 196: Same question as before, should we not set this and fallback on the default (latest)?

functional-tests/build.sbt
Aryex: Previously we had a dependency on Haddoop but it is now removed. Naming it `hadoopVersion` makes sense too since `hadoop-aws` should match with hadoop version. However, since we don't have hadoop dependency anymore I think `hadoopAwsVersion` is more appropriate.

connector/src/main/scala/com/vertica/spark/datasource/json/VerticaJsonScan.scala
Aryex: updated.

.github/workflows/on-pull-request.yml
Aryex: My thought was that I wanted to make the test configuration clear and stable since we don't want them failing on a new unsupported version, especially since this is a required PR check.

Aryex: That said, we currently don't have a good way to check when a new update comes out.

.github/workflows/nightly.yml
```diff
@@ -20,19 +19,326 @@ jobs:
...
+    needs: [run-spark-integration-tests,
+            run-spark-integration-tests-json,
+            run-spark-integration-tests-vertica-10,
+            run-integration-tests-gcs,
+            run-integration-tests-s3,
+            run-remote-tests]
+    if: failure()
     steps:
       - name: Slack Workflow Notification
         uses: Gamesight/slack-workflow-status@master
```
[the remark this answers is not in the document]
Aryex: Updated. Note that this notification only runs if the workflow failed.

.github/workflows/nightly.yml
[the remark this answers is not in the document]
Aryex: > How do we ensure we test with the desired version of Hadoop with each version of Spark (i.e. the version of Hadoop that is typically bundled with that version of Spark if you were to download them from the Spark website)?

We are pulling spark from maven, which defines its own hadoop dependency, which seems to be using hadoop 3.x. I have tried to override it but doesn't seems to work. The other alternative is to set up a cluster which I haven't tried. FYI right now our docker images are set up with Spark 3.1.2 with Hadoop 3.3.0

functional-tests/build.sbt
jeremyp-bq: But right now it is the env var name on this line, so it should be `hadoopAwsVersion` if that is what we're going with.

Aryex: OH i see what you mean. Should be fixed with next push.

.github/workflows/nightly.yml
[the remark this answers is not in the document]
Aryex: I think the simplest might be to set up a cluster instead of using sbt. The reason is Spark 3.0 to 3.3 on maven uses different Hadoop dependencies. If we want to override the Hadoop version, we would need to override the correct dependency depending on the Spark and Hadoop version we want. But I guess more importantly I wasn't able to figure out which dependency to overwrite for Spark 3.2 + vs Hadoop 2.7. 
Note that previously, we had a `"org.apache.hadoop" % "hadoop-hdfs" % "x.x.x"` dependency, which only worked for Spark 3.0.0.

.github/workflows/nightly.yml
jeremyp-bq: Also should this be HADOOP_AWS_VERSION now?

Aryex: I think these we can use the defaults

.github/workflows/nightly.yml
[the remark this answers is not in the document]
jeremyp-bq: Can we add hadoop-hdfs back to the build.sbt and use the example I described above to specify the exact version of Spark and Hadoop we want to test with in each case (which we know are compatible with each other)?

For Spark 3.3.0 that should be Hadoop 3.3.x (e.g. 3.3.2).

functional-tests/build.sbt
```diff
@@ -25,8 +25,15 @@ name := "spark-vertica-connector-functional-tests"
...
-val sparkVersion = Option(System.getProperty("sparkVersion")).getOrElse("3.3.0")
-val hadoopVersion = Option(System.getProperty("hadoopVersion")).getOrElse("3.3.0")
+val sparkVersion = Option(System.getProperty("sparkVersion")) match {
+  case Some(sparkVersion) => sparkVersion
+  case None => sys.env.getOrElse("SPARK_VERSION", "[3.3.0, 3.4.0)")
+}
+
+val hadoopVersion = Option(System.getProperty("hadoopVersion")) match {
+  case Some(hadoopAws) => hadoopAws
+  case None => sys.env.getOrElse("HADOOP_VERSION", "3.3.2")
```
jeremyp-bq, line 35: Does it make sense to get the latest 3.3 version here (`[3.3.0, 3.4.0)`)?

Should probably also make sure the connector build.sbt uses the same versions we are testing here (even if the values are hard-coded there and don't get updated as often).

functional-tests/build.sbt
```diff
@@ -25,8 +25,15 @@ name := "spark-vertica-connector-functional-tests"
...
 
-val sparkVersion = Option(System.getProperty("sparkVersion")).getOrElse("3.3.0")
-val hadoopVersion = Option(System.getProperty("hadoopVersion")).getOrElse("3.3.0")
+val sparkVersion = Option(System.getProperty("sparkVersion")) match {
+  case Some(sparkVersion) => sparkVersion
+  case None => sys.env.getOrElse("SPARK_VERSION", "[3.3.0, 3.4.0)")
+}
+
+val hadoopVersion = Option(System.getProperty("hadoopVersion")) match {
+  case Some(hadoopAws) => hadoopAws
```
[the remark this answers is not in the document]
Aryex: updated

functional-tests/build.sbt
Aryex: updated.

Aryex: updated back to static hadoop version since it was failing with Spark 3.1 vs Hadoop [3.3.1, 3.4.0)

jeremyp-bq (approved):

Issue #380, opened 2022-04-28
[BUG] Vertica Spark Connector cannot run on EMR


aramakrishnan19:
## Environment
- Spark version: Spark 3.1.2
- Hadoop version: Amazon 3.2.1 
- Vertica version: 11.1.
- Vertica Spark Connector version: 3.1.0
- Java version: openjdk version "1.8.0_332"
- Additional Environment Information: Running on EMR (emr-6.5.0)

---

## Problem Description
- Describe the issue in as much details as possible, so it is possible to reproduce it.

Vertica Spark Connector fails to run on EMR because the spark version resolves to `3.1.2-amzn-1`

The below code fails to parse the version (fails to convert the semantic version [patch] to integer), as a result considers that it is running on latest version and try to load (class=org.apache.spark.sql.connector.read.SupportsPushDownAggregates) which isnt avialable in 3.1.2 version.

https://github.com/vertica/spark-connector/blob/main/connector/src/main/scala/com/vertica/spark/datasource/v2/VerticaDatasourceV2Table.scala
```
var sparkNewerThan31 = true
        try {
          val sparkVersion = SparkSession.getActiveSession.get.version
          val versionList = sparkVersion.split("\\.").map(_.toInt)
          if(versionList.length >= 2 && versionList(0) == 3 && versionList(1) < 2) {
            sparkNewerThan31 = false
          }
        }
        catch { // Couldn't recgonize version string, assume newer version
          case _: java.lang.NumberFormatException => sparkNewerThan31 = true
        }
```
1. Steps to reproduce:
2. Expected behaviour:
3. Actual behaviour:
4. Error message/stack trace:
5. Code sample or example on how to reproduce the issue:

---

## Spark Connector Logs
- Add related logs entries here.

```java.lang.NoClassDefFoundError: org/apache/spark/sql/connector/read/SupportsPushDownAggregates
  at java.lang.ClassLoader.defineClass1(Native Method)
  at java.lang.ClassLoader.defineClass(ClassLoader.java:756)
  at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
  at java.net.URLClassLoader.defineClass(URLClassLoader.java:473)
  at java.net.URLClassLoader.access$100(URLClassLoader.java:74)
  at java.net.URLClassLoader$1.run(URLClassLoader.java:369)
  at java.net.URLClassLoader$1.run(URLClassLoader.java:363)
  at java.security.AccessController.doPrivileged(Native Method)
[...]

PR #458, opened 2022-07-08
Refactor reflections usage and making VerticaVersion general

Aryex:
### Summary

This PR refactored usages of reflections into a separate class and VerticaVersion to a more generic Version class.

### Description

- Added `ReflectionTools` containing API to invoke reflection. It is ignored from test coverage.

- `VerticaVersion` was refactored to the more generic `Version` 

- Added `SparkVersionUtils` for getting Spark version and compatibility methods. Compatibility logic can now be tested.

- `VerticaVersionUtil` refactored to be consistent with `SparkVersionUtils`. In particular,  `getVersion` now returns an Either and added a new method `getVersionOrDefault`

For reviewing, you can start at `VerticaDataSourceV2Read` and `VerticaDataSourceV2Table`

### Additional Reviewers
@jeremyp-bq 
@alexr-bq 
@jonathanl-bq 
@alexey-temnikov


connector/src/main/scala/com/vertica/spark/util/spark/SparkUtils.scala
```diff
@@ -11,25 +11,23 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package com.vertica.spark.util.version
+package com.vertica.spark.util.spark
 
+import com.vertica.spark.util.version.Version
 import org.apache.spark.sql.SparkSession
 
-object SparkVersionUtils {
```
jeremyp-bq, line 18: I think this should have the same naming convention and be in the same place as the Vertica and other version-related classes (e.g. `vertica/spark/util/version/SparkVersionUtils.scala`).

connector/src/main/scala/com/vertica/spark/util/compatibilities/DSReadCompatibilityTools.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import com.vertica.spark.util.reflections.ReflectionTools
+import com.vertica.spark.util.version.Version
+import org.apache.spark.sql.connector.expressions.aggregate.Aggregation
+import org.apache.spark.sql.connector.expressions.Expression
+
+/**
+ * Contains utility methods to support backward compatibility
+ * */
+class DSReadCompatibilityTools(reflection: ReflectionTools = new ReflectionTools) {
```
jeremyp-bq, line 24: Why create multiple compatibility classes for different Spark capabilities, while having all the capabilities for Vertica in the VerticaVersionUtils class (with the various check* methods)?

Do we plan on extracting the Vertica compatibility features into different classes as well?  Or would it be better to keep all Spark capability checks in one class like VerticaVersionUtils?  I have no preference, just want to see consistency between the Spark and Vertica version handling.

connector/src/main/scala/com/vertica/spark/util/version/Version.scala
```diff
@@ -0,0 +1,37 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.vertica.spark.util.version
+
+/**
+ * A class representing a version string of format major.minor.patch-hotfix.
+ * Only digits are allowed for minor, patch, and hotfix.
+ * */
+case class Version(major: Int, minor: Int = 0, servicePack: Int = 0, hotfix: Int = 0) extends Ordered[Version] {
```
jeremyp-bq, line 20: There was a case where the third ordinal of the Spark version in Amazon EMR was not a number (see #380).  So I think this class needs to handle cases where some of these values are not numbers.

connector/src/main/scala/com/vertica/spark/util/spark/SparkUtils.scala
Aryex: My thought was that `SparkUtils` would contain not just version-related stuff but helper methods for SparkSession. That said, I can't think of anything else to grab from the session other than the version string so maybe it should be moved to the version package and renamed.

connector/src/main/scala/com/vertica/spark/datasource/v2/VerticaDatasourceV2Read.scala
```diff
@@ -177,20 +180,9 @@ class VerticaScanBuilderWithPushdown(config: ReadConfig, readConfigSetup: DSConf
...
-        .getDeclaredMethod("groupByColumns")
-        .invoke(aggregation)
-        .asInstanceOf[Array[Expression]]
-      // $COVERAGE-ON$
-    } else {
-      aggregation.groupByExpressions()
-    }
-
-    groupByExpressions
+    val sparkVersion = SparkUtils.getVersion(SparkUtils.getVersionString).getOrElse(Version(3,3))
```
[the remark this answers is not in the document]
Aryex: Since I expect `getVersion` to be used in different places (It is also used in `VerticaDataSourceV2Table` as well), it doesn't return a default version so to be clear on when a version string was not able to be inferred and the caller should handle it.

This is inconsistent with the VerticaVersion which I wanted to refactor to not return a default value but I thought it was a bit out of the scope of this PR. 

That said, I do think I should extract the hard-coded versions to a named constant for clarity.

connector/src/main/scala/com/vertica/spark/util/compatibilities/DSReadCompatibilityTools.scala
Aryex: My thought was that the `VerticaVersionUtils` methods were more of a "check" method returning true/false, whereas these methods are returning compatible objects.

But do I feel like that's just being pedantic? They are all methods ensuring compatibilities.

connector/src/main/scala/com/vertica/spark/util/version/Version.scala
Aryex: Yes, I had considered it. Currently, the regex would only pick up the first three numbers separated by the dots and ignored the rest. This is because we currently only care about the Spark version number, but I guess it doesn't harm to record the extra information.

jeremyp-bq: I guess it's fine then, just wanted to make sure this would not cause a regression for EMR.

Aryex: I should not. I've included a unit test for that case as well
.

connector/src/main/scala/com/vertica/spark/util/version/VerticaVersionUtils.scala
```diff
@@ -25,16 +25,16 @@ import scala.util.Try
 object VerticaVersionUtils {
 
   private val logger = LogProvider.getLogger(this.getClass)
-  private val version: Option[VerticaVersion] = None
+  private val version: Option[Version] = None
   // Should always be the latest major release.
   // scalastyle:off
-  val VERRTICA_LATEST: VerticaVersion = VerticaVersion(11)
+  val VERRTICA_LATEST: Version = Version(11)
```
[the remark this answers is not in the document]
Aryex: Fixed

connector/src/main/scala/com/vertica/spark/util/compatibilities/DSReadCompatibilityTools.scala
Aryex: `SparkVersionUtils` now contains these compatibility methods.

connector/src/main/scala/com/vertica/spark/datasource/v2/VerticaDatasourceV2Read.scala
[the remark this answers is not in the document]
Aryex: Extracted the versions to constants. I couldn't think of a good name for them though so suggestions are welcome.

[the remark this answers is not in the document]
Aryex: Also refactored `VerticaVersionUtil` to be consistent, with `getVersion `returning a `ConnectorResult` instead. Added a new method `getVersionOrDefault`

connector/src/main/scala/com/vertica/spark/datasource/core/factory/VerticaPipeFactory.scala
```diff
@@ -75,7 +74,7 @@ object VerticaPipeFactory extends VerticaPipeFactoryInterface {
           case Some(session) => Some(session.sparkContext)
         }
 
-        val verticaVersion = if(getVersion) VerticaVersionUtils.getVersion(readLayerJdbc.get) else VerticaVersion(0)
+        val verticaVersion = if(getVersion) VerticaVersionUtils.getVersionOrDefault(readLayerJdbc.get) else Version(0)
         val schemaTools = if (verticaVersion.major == 10) new SchemaToolsV10 else new SchemaTools
```
jeremyp-bq, line 78: I see here and at line 99 (and perhaps other places) we should probably use lesserOrEqual.  And/or we should specify what minimum version of Vertica we support in the README (we only have tests for 10.x+).

jeremyp-bq: Although that might cause issues with the `Version(0)` part, should we default to `Version(VerticaVersionUtils.VERTICA_DEFAULT)`?

connector/src/main/scala/com/vertica/spark/datasource/v2/VerticaDatasourceV2Table.scala
```diff
@@ -84,28 +86,8 @@ class VerticaTable(caseInsensitiveStringMap: CaseInsensitiveStringMap, readSetup
...
-          classOf[VerticaScanBuilderWithPushdown]
-            .getDeclaredConstructor(classOf[ReadConfig], classOf[DSConfigSetupInterface[ReadConfig]])
-            .newInstance(config, readSetupInterface)
-        }
-        else {
-          classOf[VerticaScanBuilder]
-            .getDeclaredConstructor(classOf[ReadConfig], classOf[DSConfigSetupInterface[ReadConfig]])
-            .newInstance(config, readSetupInterface)
-        }
+        val sparkVersion = sparkVersionTools.getVersion.getOrElse(Version(3,2))
```
jeremyp-bq, line 89: Same thing here, can we make use of the constants on SparkVersionTools to get the latest version?

connector/src/main/scala/com/vertica/spark/datasource/core/factory/VerticaPipeFactory.scala
Aryex: Readme now mentioned we have only tests against vertica 10 or higher.

PR #476, opened 2022-08-09
Fix create table statement string parsing

Aryex:
### Summary

Fixes #475 

`SchemaTools.inferExternalTableSchema` did not account for nested parenthesis in create-external-table string, thus failing. It should now be able to handle such cases. 

Added `StringParsingUtils` that contains general purpose parsing methods. For reviewing, start at `SchemaTools.inferExternalTableSchema()`

Tests are updated to include the failing cases.

### Related Issue

Close #475 

### Additional Reviewers

@alexey-temnikov 
@jonathanl-bq 
@jeremyp-bq 
@alexr-bq

PR #480, opened 2022-08-15
Release 3.3.2

Aryex:
### Summary

Patch release 3.3.2, include fixes for #476 

### Additional Reviewers

@alexey-temnikov 
@jeremyp-bq 
@alexr-bq 
@jonathanl-bq

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 [![license](https://img.shields.io/badge/license-Apache%202.0-orange.svg)](https://opensource.org/licenses/Apache-2.0)
 
 ![vertica-tested](https://img.shields.io/badge/Vertica%20Tested-10%20%7C%2011%20%7C%2012-blue)
-![spark-tested](https://img.shields.io/badge/Spark%20tested-3.0%20%7C%203.1%20%7C%203.2%20%7C%203.3-blue)
+![spark-tested](https://img.shields.io/badge/Spark%20Tested-3.0%20%7C%203.1%20%7C%203.2%20%7C%203.3-blue)
 
 This component acts as a bridge between Spark and Vertica, allowing the user to either retrieve data from Vertica for processing in Spark, or store processed data from Spark into Vertica.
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -50,6 +53,11 @@ The methods for creating a spark cluster are documented [here](https://spark.apa
 
 Once you have a spark cluster, you can run such an application with spark-submit, including the connector jar.
 
+#### Spark compatibility
+
+The connector supports all Spark 3 releases. Our release numbers is of the major.minor.patch format, with the major 
+and minor numbers always matches that of the latest Spark 3 release.
+
```
jeremyp-bq, line 60: I don't think we need this as there is a section just above that states which versions we support and also in CONTRIBUTING.md.  Also, this seems to break up the sentence just above with the code snippet below.

README.md
```diff
@@ -10,6 +10,9 @@
 [![contributing](https://img.shields.io/badge/contributing-read-orange)](CONTRIBUTING.md)
 [![license](https://img.shields.io/badge/license-Apache%202.0-orange.svg)](https://opensource.org/licenses/Apache-2.0)
 
+![vertica-tested](https://img.shields.io/badge/Vertica%20Tested-10%20%7C%2011%7C%2012-blue)
+![spark-tested](https://img.shields.io/badge/Spark%20Tested-%3E%3D%203.0.0-blue)
```
jeremyp-bq, line 14: Need to be consistent with the versions and spacing (missing a space after the 11 in the Vertica versions).  I think the "|"-separated list of versions looks nicer and easier for customers to spot the version they use, but if we do use the ">=" instead we should do it for both Vertica and Spark versions.

README.md
Aryex: Removed. Versioning information is now only in contributing.

README.md
Aryex: updated to list out spark 3.0 3.1 etc

jeremyp-bq (approved):

PR #483, opened 2022-08-16
Update scala examples

Aryex:
### Summary

Some fixes to the scala example folder:
- Added a script for Kerberos example
- Fixed `submit-example.sh` with the correct jar name
- updated readme with procedure for using a modified connector jar

### Additional Reviewers
@jeremyp-bq 
@alexey-temnikov 
@alexr-bq 
@jonathanl-bq


examples/scala/src/main/scala/example/examples/BasicReadWriteExamples.scala
```diff
@@ -49,7 +49,7 @@ class BasicReadWriteExamples(spark: SparkSession) {
   /**
    * A simple example demonstrating how to write into Vertica, then read back the data in a dataframe.
    * */
-  def writeThenRead(): Unit = {
+  def writeThenReadWithHDFS(): Unit = {
```
jeremyp-bq, line 52: I think it is OK to leave it as writeThenRead as HDFS/Hadoop is the default case.  If you do rename it there are references in the README that also need to be updated.

Aryex: updated.

examples/scala/src/main/scala/example/Main.scala
```diff
@@ -22,6 +22,7 @@ object Main {
 
     // Define a Spark master here
     val spark = SparkSession.builder()
+      // .master("local[*]")
```
jeremyp-bq, line 25: Should we uncomment this or remove it?

Aryex: removed.

jeremyp-bq (approved):

Issue #484, opened 2022-08-19
Flakey CI Tests


Aryex:
# Descriptions
Our nightly tests and weekly tests would randomly fails when trying to do a docker-compose up. This seems to always happen to our jobs using a matrix, and we have only observed one run failing, but no more. The failure seems to always happen with hdfs container, and the error is from "a port is not available". 

Re-running the failed tests and it would pass. We are unsure if this has something to do with running too many parallel jobs using matrices.

jeremyp-bq:
Last failure was under the Weekly Tests was on September 2.  The failing test was the second in the matrix (`Vertica ([3.0.0, 3.1.0), 2.7.4, 12.0.0-0)`), once again complaining about the HDFS port 50020 not being available.

Will continue to monitor this issue to see if the additional `docker compose down` helps.

PR #485, opened 2022-08-19
Shutdown container before docker-compose up

Aryex:
### Summary

Added a docker-compose down in the hope that it properly closes the ports and allow for the following docker-compose up to succeed always

### Related Issue

#484 

### Additional Reviewers
@alexey-temnikov 
@jeremyp-bq 
@jonathanl-bq 
@alexr-bq

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -37,7 +37,7 @@ jobs:
       - name: Shutdown container first
         run: cd docker && docker-compose down
       - name: Run docker compose
-        run: docker-compose up -d
+        run: cd docker && docker-compose up -d
         env:
           SPARK_VERSION: ${{matrix.runtime.spark}}
           HADOOP_VERSION: ${{matrix.runtime.hadoop}}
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yml
```diff
@@ -34,6 +34,8 @@ jobs:
     steps:
       - name: Checkout the project
         uses: actions/checkout@v2
+      - name: Shutdown container first
+        run: cd docker && docker-compose down
```
jeremyp-bq, line 38: It may take weeks before we know if this helps or not, I think we saw the issue around once a week in the past.  This may also affect the weekly tests since we have matrix tests there as well.

Aryex: Yeah, I thought that this was a small change so wanted to try and monitor it going forward. If it does fix it then we could add it to weekly as well.

alexr-bq (approved):

jeremyp-bq (approved):

Issue #507, opened 2022-10-05
Support table truncate when writing in overwrite mode


jmyrberg:
## Is your feature request related to a problem? Please describe.

We have a table where all the rows need to be re-written after a Spark batch job. The table has existing permissions that should be preserved. However, the table permissions are lost when the rows are written in *overwrite* mode:

```python
(sdf.write.format('com.vertica.spark.datasource.VerticaSource')
    .mode('overwrite').options(**options).save())
```

## Describe the solution you'd like

In *overwrite* mode, there should be an option to truncate the target table instead of dropping it before re-write.

The standard Spark JDBC connector allows one to set `truncate` option to solve this ([link](https://spark.apache.org/docs/latest/sql-data-sources-jdbc.html)).

## Describe alternatives you've considered

Workarounds exist, but they are not very practical:
- Execute truncate statement through JDBC using some other library and write the rows using the connector in *append* mode
- Pass the privileges through `target_table_sql` parameter
- Using some external script/procedure to re-set the privileges

## Additional context

What the solution could look like after implementation:

```python
options['truncate'] = True

(sdf.write.format('com.vertica.spark.datasource.VerticaSource')
    .mode('overwrite').options(**options).save())
```

alexey-temnikov:
Hi @jmyrberg, thank you for raising this. I agree with your suggestion, we will look into this enhancement.

jmyrberg:
> @jmyrberg, the `truncate` option (see the documentation in `README.md`) has been added as part of release 3.3.4, which is now available in Maven.

Awesome, we’ll have a look, thank you! 👍

PR #509, opened 2022-10-19
added option for truncation in overwrite

ai-bq:
### Summary

Some users require the option of truncating a table rather than dropping it in overwrite mode to retain table privileges. 

### Description

Truncate table was added to table utilities as well as the option of truncation through the dataframe writing options.

### Related Issue

#507 

### Additional Reviewers

@alexey-temnikov 
@jeremyp-bq 
@jonathanl-bq 
@alexr-bq

### DIFF
--- a/connector/src/main/scala/com/vertica/spark/config/WriteConfig.scala
+++ b/connector/src/main/scala/com/vertica/spark/config/WriteConfig.scala
@@ -89,7 +89,8 @@ final case class DistributedFilesystemWriteConfig(jdbcConfig: JDBCConfig,
                                                   saveJobStatusTable: Boolean,
                                                   mergeKey: Option[ValidColumnList] = None,
                                                   timeOperations : Boolean = true,
-                                                  arrayLength: Long = 0
+                                                  arrayLength: Long = 0,
+                                                  truncate: Boolean
                                                  ) extends WriteConfig {
   private var overwrite: Boolean = false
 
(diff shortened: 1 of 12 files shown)


jeremyp-bq:
@ai-bq you'll need to sign the CLA (you should have gotten an email) before that check will pass.

Also note there is a large file at `examples/scala/cs` that shouldn't be committed.

connector/src/main/scala/com/vertica/spark/config/WriteConfig.scala
```diff
@@ -89,7 +89,8 @@ final case class DistributedFilesystemWriteConfig(jdbcConfig: JDBCConfig,
                                                   saveJobStatusTable: Boolean,
                                                   mergeKey: Option[ValidColumnList] = None,
                                                   timeOperations : Boolean = true,
-                                                  arrayLength: Long = 0
+                                                  arrayLength: Long = 0,
+                                                  truncate: Boolean
```
jeremyp-bq, line 93: We should add this to the Javadoc above (and add the other missing params at the same time).

At some point, although not in this PR, we should consider just passing an options object rather than adding more and more parameters.

connector/src/main/scala/com/vertica/spark/util/table/TableUtils.scala
```diff
@@ -260,6 +265,11 @@ class TableUtils(schemaTools: SchemaToolsInterface, jdbcLayer: JdbcLayerInterfac
       .left.map(err => err.context("JDBC Error dropping table"))
   }
 
+  def truncateTable(tablename: TableName): ConnectorResult[Unit] = {
```
jeremyp-bq, line 268: Not sure if we can add a unit test for this as well, we should follow whatever unit testing was done for drop table.

PR #510, opened 2022-10-20
Option for truncation in overwrite mode

### LINKED ISSUE #507: Support table truncate when writing in overwrite mode

ai-bq:
### Summary

Some users require the option of truncating a table rather than dropping it in overwrite mode to retain table privileges.

### Description

Truncate table was added to table utilities as well as the option of truncation through the dataframe writing options.

### Related Issue

Closes #507 

### Additional Reviewers

@alexey-temnikov 
@alexr-bq 
@jonathanl-bq 
@jeremyp-bq


connector/src/test/scala/com/vertica/spark/util/table/TableUtilsTest.scala
```diff
@@ -383,6 +383,19 @@ class TableUtilsTest extends AnyFlatSpec with BeforeAndAfterAll with MockFactory
     checkResult(utils.dropTable(TableName(tablename, None)))
   }
 
+    it should "Truncate a table" in {
```
jeremyp-bq, line 386: nit: wrong indentation.

functional-tests/src/main/scala/com/vertica/spark/functests/endtoend/EndToEndTests.scala
```diff
@@ -1103,6 +1103,48 @@ class EndToEndTests(readOpts: Map[String, String], writeOpts: Map[String, String
...
+    catch{
+      case err : Exception => fail(err)
+    }
+    finally {
+      stmt.execute("DROP USER ALEX")
+      stmt.close()
+    }
+
+    TestUtils.dropTable(conn, tableName)
+
```
jeremyp-bq, line 1145: nit: extra line here and a couple more above.

jeremyp-bq (approved):

PR #517, opened 2022-11-15
Add parquet export support for complex types as of Vertica 12.0.2

### LINKED ISSUE #499: Support complex types via export to parquet

We currently must export to JSON in order to support Vertica complex types.  However, in the next release of Vertica (12.0.2) it is expected that the export to parquet functionality will support complex types.

In order to avoid exporting to JSON in some cases as we currently do, it will be more consistent and faster to always export to parquet (provided it supports all complex types).  Although we will likely need to keep the current code to support the versions of Vertica where complex types were only supported with export to JSON.

Will need to do some advanced testing once the next Vertica release is available in order to confirm support of complex types when using export to parquet.

ai-bq:
### Summary

Vertica 12.0.2 is adding full support of Parquet export for complex types. Spark Connector needed to be updated in a way that

1. JSON Export is no longer forced when a complex type column is registered on initial scan
2. Any user who is on a prior version of Vertica still has this JSON Export as the default case

### Description

The VerticaScanBuilder runs prior to the Spark read pipe and includes a JSON check which returns true if complex columns are found in the table schema. This function remains the same with the exception of a conditional that checks the version of Vertica (>12.0.2 does not need this complex type check).

To successfully fetch the version of Vertica, it is added as a param of VerticaReadMetadata and passed along when VerticaScanBuilder fetches Vertica metadata in the JSON check.

### Related Issue

Closes #499. 

### Additional Reviewers

@alexey-temnikov 
@alexr-bq 
@jonathanl-bq 
@jeremyp-bq


connector/src/main/scala/com/vertica/spark/datasource/v2/VerticaDatasourceV2Read.scala
```diff
@@ -92,14 +94,18 @@ class VerticaScanBuilder(config: ReadConfig, readConfigSetup: DSConfigSetupInter
...
+          case (Right(metadata), requiredSchema) =>
             val schema: StructType = if (requiredSchema.nonEmpty) {
               requiredSchema
             } else {
-              metadataSchema
+              metadata.schema
             }
-            config.useJson || ctTools.filterComplexTypeColumns(schema).nonEmpty
+            if(config.useJson) true
+            else if(metadata.version < VerticaVersionUtils.VERTICA_12_0_2) {
```
jeremyp-bq, line 105: I'm wondering if we should put this logic under a method in VerticaVersionUtils (e.g. VerticaVersionUtils.checkComplexTypesParquetExport(version: Version)).

Unfortunately, I see there is inconsistency between using the VersionUtils class to determine functionality, and only using the version numbers defined in that class with logic in other classes.

connector/src/main/scala/com/vertica/spark/datasource/core/DSConfigSetup.scala
```diff
@@ -705,6 +709,16 @@ class DSReadConfigSetup(val pipeFactory: VerticaPipeFactoryInterface = VerticaPi
         }
     }
   }
+
+  override def getTableMeta(config: ReadConfig): ConnectorResult[VerticaReadMetadata] =  {
```
jeremyp-bq, line 713: We discussed offline that we could probably refactor getTableSchema and getTableMeta into a single method since they are mostly identical, then have the calling class get what they need from the returned metadata object (schema and/or version).

connector/src/main/scala/com/vertica/spark/datasource/v2/VerticaDatasourceV2Read.scala
```diff
@@ -92,14 +94,18 @@ class VerticaScanBuilder(config: ReadConfig, readConfigSetup: DSConfigSetupInter
...
+        (readConfigSetup.getTableMeta(config), config.getRequiredSchema) match {
+          case (Right(metadata), requiredSchema) =>
             val schema: StructType = if (requiredSchema.nonEmpty) {
               requiredSchema
             } else {
-              metadataSchema
+              metadata.schema
             }
-            config.useJson || ctTools.filterComplexTypeColumns(schema).nonEmpty
+            if(config.useJson) true
```
jeremyp-bq, line 104: This is what is causing the build to fail, need to add braces around if statements to satisfy the scalastyle.

connector/src/main/scala/com/vertica/spark/datasource/core/DSConfigSetup.scala
```diff
@@ -49,9 +50,9 @@ trait DSConfigSetupInterface[T] {
   def performInitialSetup(config: T): ConnectorResult[Option[PartitionInfo]]
 
   /**
-   * Returns the schema for the table as required by Spark.
+   * Returns the metadata for the table as required by Spark.
    */
-  def getTableSchema(config: T): ConnectorResult[StructType]
+  def getTableMeta(config: T): ConnectorResult[VerticaReadMetadata]
```
jeremyp-bq, line 55: nit: Rename to getTableMetadata as we refer to it as metadata, not meta, in all other functions/variables.

connector/src/main/scala/com/vertica/spark/util/version/VerticaVersionUtils.scala
```diff
@@ -84,6 +86,18 @@ object VerticaVersionUtils {
     }
   }
 
+  def checkComplexTypesParquetExport(schema: StructType, version: Version): ConnectorResult[Unit] = {
+    val (nativeCols, complexTypeCols) = complexTypeUtils.filterColumnTypes(schema)
+    val complexTypeFound = complexTypeCols.nonEmpty
```
jeremyp-bq, line 91: complexTypeFound is not used in this function.

Also, this function is for export from Vertica, so should the check below be for ComplexTypeReadNotSupported?

connector/src/main/scala/com/vertica/spark/datasource/core/DSConfigSetup.scala
```diff
@@ -671,6 +672,7 @@ class DSReadConfigSetup(val pipeFactory: VerticaPipeFactoryInterface = VerticaPi
           case _ => MissingMetadata().invalidNec
         }
       }
+
```
jonathanl-bq, line 675: nit: try to be careful about adding additional whitespace to files, since it creates noise for the reviewers

connector/src/main/scala/com/vertica/spark/datasource/core/DSConfigSetup.scala
```diff
@@ -691,17 +693,18 @@ class DSReadConfigSetup(val pipeFactory: VerticaPipeFactoryInterface = VerticaPi
   }
 
   /**
-   * Returns the schema of the table being read
+   * Returns the metadata of the table being read
    *
-   * @param config Configuration data for the read operation. Contains the metadata required for returning the table schema.
-   * @return The table schema or an error that occured trying to retrieve it
+   * @param config Configuration data for the read operation.
+   * @param version Contains the version number for Vertica.
```
jonathanl-bq, line 699: The doc comment is out of sync with the method parameters again

connector/src/main/scala/com/vertica/spark/datasource/core/DSConfigSetup.scala
```diff
@@ -763,11 +766,5 @@ class DSWriteConfigSetup(val schema: Option[StructType], val pipeFactory: Vertic
...
   }
 
-  /**
-   * Returns the same schema that was passed in to this class.
-   */
-  override def getTableSchema(config: WriteConfig): ConnectorResult[StructType] = this.schema match {
-    case Some(schema) => Right(schema)
-    case None => Left(SchemaDiscoveryError())
-  }
+  override def getTableMeta(config: WriteConfig): ConnectorResult[VerticaReadMetadata] = Right(VerticaReadMetadata(new StructType(), VerticaVersionUtils.VERTICA_DEFAULT))
```
jonathanl-bq, line 769: If we always return an empty schema, we could change this method to not return a ConnectorResult to indicate that it is infallible. More importantly, we should not be returning **READ** metadata from a **WRITE** method. It's just a naming thing, but I would be more comfortable if it was "VerticaWriteMetadata" or if it can be returned in both Read and Write cases, just "VerticaMetadata".

connector/src/main/scala/com/vertica/spark/datasource/core/VerticaDistributedFilesystemWritePipe.scala
```diff
@@ -174,6 +174,7 @@ class VerticaDistributedFilesystemWritePipe(val config: DistributedFilesystemWri
   val timer = new Timer(config.timeOperations, logger, "Writing Partition.")
 
   def startPartitionWrite(uniqueId: String): ConnectorResult[Unit] = {
+
```
jonathanl-bq, line 177: nit: extra newline

connector/src/main/scala/com/vertica/spark/datasource/v2/VerticaDatasourceV2Read.scala
```diff
@@ -92,14 +94,17 @@ class VerticaScanBuilder(config: ReadConfig, readConfigSetup: DSConfigSetupInter
...
+          case (Right(metadata), requiredSchema) =>
             val schema: StructType = if (requiredSchema.nonEmpty) {
               requiredSchema
             } else {
-              metadataSchema
+              metadata.schema
             }
-            config.useJson || ctTools.filterComplexTypeColumns(schema).nonEmpty
+            config.useJson || ((VerticaVersionUtils.checkComplexTypesParquetExport(schema, metadata.version)) match {
+                case Left(_) => ctTools.filterComplexTypeColumns(schema).nonEmpty
```
jonathanl-bq, line 105: Since we're catching an error here, might it make sense to log something to indicate that we're catching an error and falling back on a different approach?

connector/src/main/scala/com/vertica/spark/util/version/VerticaVersionUtils.scala
```diff
@@ -84,6 +86,18 @@ object VerticaVersionUtils {
...
+    if (version.lessThan(VERTICA_12_0_2)) {
```
jonathanl-bq, line 92: nit: This makes me think we should have just implemented an Ord instance on Version so we could just use the regular < operator here instead of defining our own lessThan method.

connector/src/test/scala/com/vertica/spark/common/TestObjects.scala
```diff
@@ -16,6 +16,7 @@ package com.vertica.spark.common
 import com.vertica.spark.config.{AWSOptions, BasicJdbcAuth, DistributedFilesystemReadConfig, DistributedFilesystemWriteConfig, FileStoreConfig, GCSOptions, JDBCConfig, JDBCTLSConfig, TableName, ValidFilePermissions, VerticaMetadata}
 import com.vertica.spark.datasource.core.{DataBlock, PartitionInfo, Require, VerticaPartition, VerticaPipeInterface, VerticaPipeReadInterface, VerticaPipeWriteInterface}
 import com.vertica.spark.util.error.ErrorHandling.ConnectorResult
+import com.vertica.spark.util.version.Version
```
jonathanl-bq, line 19: Why is this imported if it isn't being used anywhere here?

jonathanl-bq (changes requested):

connector/src/main/scala/com/vertica/spark/util/version/VerticaVersionUtils.scala
ai-bq: I think that's a good idea but I'm gonna leave that for another ticket as it could possibly lead to more refactoring on this one

jeremyp-bq: BTW, was just looking and the Version class and it already extends Ordered.  So we can use the various logical operators without any changes, not sure why the string operators were added.

But agree, it should be done as a separate ticket to clean that up.

jeremyp-bq (approved):

jonathanl-bq (approved):

Issue #514, opened 2022-11-04
Update tests for Vertica 12.0.2


ai-bq:
Workflows need to be updated for Vertica 12.0.2 to properly test the new addition of Parquet export for complex types.

Related to #513.

jeremyp-bq:
Closed by #526.

PR #526, opened 2022-12-08
Tests updated to include Vertica 12.0.2

### LINKED ISSUE #513: Ensure latest patch version of Vertica is being tested

Our automated tests for Vertica 12 are all for 12.0.0, but we should be testing against the latest patch version that most customers would be using.  Right now that is 12.0.1, and soon to be 12.0.2.

Update the main, daily, and weekly tests to test on 12.0.1 (and later 12.0.2) instead of 12.0.0.

ai-bq:
### Summary

Updating the tests to run on the latest version of Vertica. Nightly also has the previous version to test complex type parquet export before and after.

### Description

Version number 12.0.0 -> 12.0.2.

### Related Issue

Closes #513 and #514.

### Additional Reviewers

@alexey-temnikov 
@alexr-bq 
@jonathanl-bq 
@jeremyp-bq

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -30,7 +30,7 @@ jobs:
                    { spark: "[3.1.0, 3.2.0)", hadoop: "3.2.0" },
                    { spark: "[3.2.0, 3.3.0)", hadoop: "3.3.1" },
                    { spark: "[3.3.0, 3.4.0)", hadoop: "3.3.2" } ]
-        vertica: ["11.1.1-2", "12.0.0-0" ]
+        vertica: ["11.1.1-2", "12.0.0-0", "12.0.2-0"]
     steps:
       - name: Checkout the project
         uses: actions/checkout@v2
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -30,7 +30,7 @@ jobs:
                    { spark: "[3.1.0, 3.2.0)", hadoop: "3.2.0" },
                    { spark: "[3.2.0, 3.3.0)", hadoop: "3.3.1" },
                    { spark: "[3.3.0, 3.4.0)", hadoop: "3.3.2" } ]
-        vertica: ["11.1.1-2", "12.0.0-0", "12.0.2-0"]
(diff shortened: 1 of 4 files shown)


.github/workflows/nightly.yml
```diff
@@ -30,7 +30,7 @@ jobs:
                    { spark: "[3.1.0, 3.2.0)", hadoop: "3.2.0" },
                    { spark: "[3.2.0, 3.3.0)", hadoop: "3.3.1" },
                    { spark: "[3.3.0, 3.4.0)", hadoop: "3.3.2" } ]
-        vertica: ["11.1.1-2", "12.0.0-0" ]
+        vertica: ["11.1.1-2", "12.0.0-0", "12.0.2-0" ]
```
jeremyp-bq, line 33: I think we just run run the latest patch version of each release.  So 11.1.1-2 and 12.0.2-0 should be sufficient.  Otherwise the number of patch version we test will get out of hand and take a long time.

.github/workflows/weekly.yml
```diff
@@ -30,7 +30,7 @@ jobs:
                    { spark: "[3.1.0, 3.2.0)", hadoop: "3.2.0" },
                    { spark: "[3.2.0, 3.3.0)", hadoop: "3.3.1" },
                    { spark: "[3.3.0, 3.4.0)", hadoop: "3.3.2" } ]
-        vertica: [ "12.0.0-0" ]
+        vertica: [ "12.0.2-0" ]
```
jeremyp-bq, line 33: I think this should also include a 11.x version, like the nightly (i.e. 11.1.1-2).

jeremyp-bq:
Make sure to also update the `.github/workflows/README.md` to reflect any changes to the workflows and versions tested.  The purpose of that README is to have a summary of what we test.

.github/workflows/weekly.yml
jeremyp-bq: Actually this is specific to JSON tests, so we need 12.x or newer.  Maybe we should just remove this vertica argument and rely on latest to avoid confusion about what should be updated.

jeremyp-bq (approved):

PR #530, opened 2023-01-05
Added a complex array and Spark ML Jupyter example

### LINKED ISSUE #436: Add Jupyter notebook examples
## Description

Jupyter with Spark ML or PySpark is a common use case for our users. We should look into best practices on sharing Jupyter notebook as part of the examples on the repo. This should include instruction on how to connect a Jupyter environment to a Spark-Vertica environment.

Should try to pull in connector dependencies from maven 

Reason: Helps with adoption and reduces time to get started with Spark connector.

ai-bq:
### Summary

A common use case for users of the connector is Spark's ML tools. Additionally, Jupyter is an easy way to run examples that allows for the user to experiment and make modifications without having to rebuild examples.

### Description

Added an example of Spark's Linear Regression algorithm as well as a complex array writing example.

### Related Issue

Closes #436.

### Additional Reviewers

@alexey-temnikov 
@alexr-bq 
@jonathanl-bq 
@jeremyp-bq


examples/jupyter/faithful.csv
```diff
@@ -0,0 +1,273 @@
+"id","eruptions","waiting"
```
jeremyp-bq, line 1: To avoid things getting messy when we add more sample data, I would put this under a data folder (e.g. `./data/faithful/faithful.csv`).

examples/jupyter/linear-regression-example.ipynb
```diff
@@ -0,0 +1,317 @@
...
+      " |-- id: integer (nullable = true)\n",
+      " |-- eruptions: double (nullable = true)\n",
+      " |-- waiting: integer (nullable = true)\n",
+      "\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Load some example data from a CSV file into a dataframe and show some if its contents\n",
+    "df = (spark.read.format(\"com.databricks.spark.csv\")\n",
```
jeremyp-bq, line 147: This doesn't work without installing databricks first. So we need to add that process to the notebook.

However, I think Pandas is more prevalent and easier to use (can't recall if it is installed by default or not, if not easy to execute a conda install command in a notebook cell).  For example:
```
import pandas as pd

df = pd.read_csv("path/to/csv")
```

jeremyp-bq: Although I see a Pandas DF can not be used to write to Spark, so would need to convert it or continue with the Databricks DF (after installing the necessary package to use Databricks).

examples/jupyter/data/faithful_testing.csv
```diff
@@ -0,0 +1,111 @@
+"id","eruptions","waiting"
```
jeremyp-bq, line 1: FYI, you can use various ML packages to split the data at runtime.  For example, [scikit-learn](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.train_test_split.html).

Although that can be a nice-to-have since you have already split the data here.

ai-bq: Yeah ideally one would use that or Spark's randomSplit(). I think by using the same split data (the two .csvs) as Vertica's Linear Regression example though might be better for a side-to-side comparison?

alexey-temnikov:
[spark-connector](https://github.com/vertica/spark-connector)/[examples](https://github.com/vertica/spark-connector/tree/main/examples)/[jupyter](https://github.com/vertica/spark-connector/tree/main/examples/jupyter)/README.md should be updated to include this example, as well as describe what this example is about.

alexey-temnikov:
Command - "docker-compose --profile jupyter up -d" will not work from [spark-connector](https://github.com/vertica/spark-connector)/[examples](https://github.com/vertica/spark-connector/tree/main/examples)/jupyter/ as there is no docker-compose file. Instructions need to be fixed.

jeremyp-bq:
Do we need to add/update the examples here based on this PR?:
https://github.com/vertica/vertica-demo/pull/155

ai-bq:
> Do we need to add/update the examples here based on this PR?: [vertica/vertica-demo#155](https://github.com/vertica/vertica-demo/pull/155)

Yes, I will compare the two and update this set.

examples/jupyter/linear-regression-example-vertica-direct.ipynb
```diff
@@ -0,0 +1,294 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "3cef4d75-e4dd-4074-9075-c9fb03df0a8f",
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "pip install vertica-python & pip install PrettyTable"
```
jeremyp-bq, line 36: This should be a `%%bash` cell.  Also applies to the VerticaPy example.

You can also install multiple packages space-delimited:
```
%%bash
pip install vertica-python PrettyTable
```

jeremyp-bq (approved):

PR #535, opened 2023-02-07
Java 8 isBlank fix + JDK 8 addition to Weekly workflow 

### LINKED ISSUE #534: [BUG] Connector fails to build with Java 8 due to unsupported method
## Environment
- Java version: 8u352
---

## Problem Description
- Describe the issue in as much details as possible, so it is possible to reproduce it.

1. Steps to reproduce: Download Java 8 that is at least update 92 or higher and switch to it to build the connector.
2. Expected behaviour: Should compile fine.
3. Actual behaviour: Compilation error.
4. Error message/stack trace:
```
[error] /Volumes/Extended/Projects/Vertica/spark-connector/connector/src/main/scala/com/vertica/spark/util/schema/SchemaTools.scala:681:27: value isBlank is not a member of String
[error]           if (column.name.isBlank) {
[error]                           ^
[error] one error found
[error] (Compile / compileIncremental) Compilation failed
```
---

It seems isBlank is not supported by Java 8. We should use an external library such as Apache Common's StringUtil so that both Java 11 and Java 8 can run isBlank.

In the future Java 8 should be tested or at least have the Java 11 compilation run a backwards compatibility build on Java 8 as well.

### LINKED ISSUE #536: Add JDK 8 to Weekly Tests
Recently the compilation of the Connector was discovered to fail when building with Java 8 as per #534.

As long as Java 8 is to be supported by the Connector we should add JDK 8 to the weekly tests to ensure the compilation works on both supported versions.

ai-bq:
### Summary

Java 8 was failing compilation of the Connector due to an unsupported method "isBlank." Additionally, Java 8 compilation is never tested and should be added to GH Actions.

### Description

Imported an Apache Commons StringUtil library to handle the isBlank method call. JDK 8 has yet to have been added to Weekly testing.

### Related Issue

Closes #534 and #536.

### Additional Reviewers

@alexey-temnikov 
@alexr-bq 
@jonathanl-bq 
@jeremyp-bq

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -9,15 +9,22 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        jdk: ['11', '8']
     steps:
       - name: Checkout the project
         uses: actions/checkout@v2
+      - uses: actions/setup-java@v3
+        with:
+          distribution: 'temurin'
+          java-version: ${{ matrix.jdk }}
       - name: Build the project
         run: cd connector && sbt package
       - name: Upload the build artifact
         uses: actions/upload-artifact@v3
         with:
-          name: build-jar-file
+          name: build-jar-file-${{ matrix.jdk }}
           path: /home/runner/work/spark-connector/spark-connector/connector/target/scala-2.12/spark-vertica-connector_2.12-*.jar
   run-spark-integration-tests-json:
     name: Run regression tests using JSON export
@@ -30,6 +37,7 @@ jobs:
                    { spark: "[3.1.0, 3.2.0)", hadoop: "3.2.0" },
                    { spark: "[3.2.0, 3.3.0)", hadoop: "3.3.1" },
                    { spark: "[3.3.0, 3.4.0)", hadoop: "3.3.2" } ]
+        jdk: ['11', '8']
     steps:
       - name: Checkout the project
         uses: actions/checkout@v2
@@ -44,7 +52,7 @@ jobs:
       - name: Download the build artifact
         uses: actions/download-artifact@v2
         with:
-          name: build-jar-file
+          name: build-jar-file-${{ matrix.jdk }}
           path: ./functional-tests/lib/
       - name: Wait for Vertica to be available
         uses: nick-invision/retry@v2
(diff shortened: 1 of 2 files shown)


connector/src/main/scala/com/vertica/spark/util/schema/SchemaTools.scala
```diff
@@ -678,7 +679,7 @@ class SchemaTools(ctTools: ComplexTypesSchemaTools = new ComplexTypesSchemaTools
     def findEmptyColumnName(fields: List[StructField]): ConnectorResult[Unit] = {
       fields.headOption match {
         case Some(column) =>
-          if (column.name.isBlank) {
+          if (StringUtils.isBlank(column.name)) {
```
alexey-temnikov, line 682: 1.  Is it covered by Unit Tests?
2. Let's add another GitHub action to build connector with Java 8, to ensure issues like this won't happen in the future.

ai-bq: 1. Code coverage is unchanged so the unit test reaches that condition
2. I've added JDK8 to our Weekly workflow in another branch but I need this merged to test it properly (otherwise it'll just fail)

.github/workflows/weekly.yml
```diff
@@ -30,6 +37,7 @@ jobs:
                    { spark: "[3.1.0, 3.2.0)", hadoop: "3.2.0" },
                    { spark: "[3.2.0, 3.3.0)", hadoop: "3.3.1" },
                    { spark: "[3.3.0, 3.4.0)", hadoop: "3.3.2" } ]
+        jdk: ['11', '8']
```
jeremyp-bq, line 40: Looking at this in more detail, I like the matrix for the build job above, but I don't think we need to run each of these existing tests on both JDK 8 and 11.  Each of these tests is for a specific case (JSON, Vertica 10.x, S3, GCS, etc).  I think we should add a new job to test JDK 8 (e.g. run-spark-integration-tests-jdk8), and leave the rest on JDK 11.

Since you ran these tests against JDK8 and they pass (other than S3/GCS concurrency issues), from this point forward just adding an extra job/test against JDK8 should be sufficient.

jeremyprime (commented):
Looks good, we can always tweak the matrix for the JDK8 test in the future if we feel this is too many or too few tests.

jeremyprime (approved):

haidernaveed:
Can we get the jar of this build ? 
I can see the latest release is of Dec 2022 and im having the **isBlank**() error in that
@jeremyprime @ai-bq @alexey-temnikov @jonathanl-bq

PR #548, opened 2023-06-09
Updated Spark logging and Docker information

jeremyprime:
### Summary

Updated Spark logging and Docker information

### Description

Updated documentation on Spark/Spark Connector logging with specific working examples.  Also updated some missing information and steps in Docker container management.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -8,21 +8,17 @@ If you want to try these examples on our Docker environment, then:
 ```sh
 git clone https://github.com/vertica/spark-connector.git
 ```
-3. Navigate to the `docker` folder:
-```sh
-cd spark-connector/docker
-```
-4. Start the appropriate configuration:
+3. Start the appropriate configuration:
 ```sh
 cd spark-connector/docker
 docker-compose up -d
-# or
+# or, for Kerberos
 docker-compose -f docker-compose-kerberos.yml up -d
 ```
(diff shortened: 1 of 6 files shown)


examples/README.md
```diff
@@ -3,21 +3,22 @@
...
 docker-compose up -d
-# or
+# or, for Kerberos
 docker-compose -f docker-compose-kerberos.yml up -d
 ```
 4. Get a shell to the client container:
 ```sh
 docker exec -it docker-client-1 bash
-# or
+# or, for Kerberos
```
jonathanl-bq, line 21: I wonder if it wouldn't be possible to just have one name instead of requiring this "or"

jonathanl-bq (approved):
I left one question about maybe simplifying one of the Docker commands, but otherwise, looks good.

examples/README.md
jeremyprime: The Kerberos configuration requires very specific hostnames, and thus container names, for the networking between the containers to work.  But it is also best to avoid generic names that might conflict with other containers on the host ("client", "spark", etc).  In other words, we could keep consistent names, but they would be dictated by what Kerberos requires (so not worth it IMO).

jonathanl-bq: Okay, I'm fine with that.

PR #565, opened 2026-06-17
JDK 21 compatibility: upgrade connector to Spark 4.1.1 and align build/API changes

vgarg-mf:
**Description:**
This MR updates the connector for JDK 21 compatibility by moving to the Spark 4.1.1 stack and applying required source/build adjustments so the project compiles cleanly with the newer APIs.

**Key changes:**
Upgraded Spark dependencies to 4.1.1 (core and SQL).
Updated Scala patch version (2.13.18) and Scapegoat plugin version (3.3.4) for build/tooling compatibility.
Added Spark 4.x Parquet config handling and explicit Parquet field-id read/write flags for cross-version behavior.
Updated V2 catalog method signature to use override where required by newer Spark interfaces.
Refactored AnalysisException construction to the newer constructor shape expected by upgraded Spark APIs.
Updated connector version metadata:
    3e87dcb set connector version to 4.4.1
    559f2ad adjusted connector version to 4.1.1 (final)

**Outcome:**
Connector codebase is aligned with Spark 4.1.1 and JDK 21 compilation requirements.
Versioning now reflects the final intended connector release value.


functional-tests/build.sbt
cbharadwajp: Functional tests are being validated on a spark 3.5.5 and the shipped connector targets 4.1.1. Tests should be in-line to the shipped artifacts.

performance-tests/build.sbt
cbharadwajp: Performance tests are being validated on a spark 3.5.5 and the shipped connector targets 4.1.1. Tests should be in-line to the shipped artifacts.

cbharadwajp (changes requested):
Requested changes

cbharadwajp:
Also please check if we are supporting backward compatibility with JDK version. If yes, document the versions upto which the backward compatibility will be supported in release notes.

functional-tests/build.sbt
vgarg-mf: Updated the reference

performance-tests/build.sbt
vgarg-mf: Updated the reference

cbharadwajp (approved):
Changes are approved.

PR #566, opened 2026-07-02
Fix cross-database JDBC cache reuse in DSv2 read/write flow

vgarg-mf:
**Description**
This MR fixes a JDBC layer reuse issue in the Vertica Spark Connector DataSource V2 path when reading from one Vertica database and writing to another in the same Spark application.

**Problem**
The connector caches read/write JDBC layers in VerticaPipeFactory.
Previously, cached layers were reused if open, without validating that the cached connection belonged to the same JDBCConfig.

In mixed source/target DB workflows, this could cause stale connection context reuse across operations, leading to schema/table lookup being executed against the wrong database after plan recomputation.

**Root Cause**
In VerticaPipeFactory, checkJdbcLayer only checked isClosed and did not validate JDBC configuration identity.

**Fix**

1. Added JDBC config equality helper in VerticaJdbcLayer:
   A. matchesConfig(other: JDBCConfig): Boolean
2. Updated VerticaPipeFactory.checkJdbcLayer logic:
   A. Reuse cached layer only when open and config matches
   B. If config differs, close old layer and create a new VerticaJdbcLayer
   C. If absent/closed, create new layer as before

**Files Changed**
   VerticaPipeFactory.scala
   VerticaJdbcLayer.scala

**Behavior Impact**

Before:
Reused any open cached JDBC layer

After:
Reuses cached JDBC layer only if open and JDBCConfig matches
Recreates layer on config mismatch

**Validation**
Validated with source and target on different Vertica DB endpoints:

1. Without intermediate action: df.write.save() succeeds
2. With intermediate action: df.show(); df.write.save() succeeds
3. Source reads/export use source DB connection
4. Write/copy/commit use target DB connection
5. No relation does not exist/schema-lookup-against-wrong-db failure observed

**Risk**
Low to moderate:
1. Narrowly scoped to JDBC cache reuse logic
2. May open new connections when config changes (intended correctness behavior)
3. No change to DSv2 lifecycle semantics, COPY/export logic, or user options


ajay-abrol2 (approved):
Code looks good to me so approve this PR.

